PR #42, opened 2024-11-27
Cypher validation fix and add lists to the attribute types

galshubeli:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new attribute type, `LIST`, enhancing the attribute options available in the system.

- **Bug Fixes**
	- Improved validation logic for relation labels in the validation functions, ensuring more accurate checks against the ontology.

- **Chores**
	- Updated project version from 0.3.2 to 0.3.3 in the configuration file.

### DIFF
--- a/graphrag_sdk/helpers.py
+++ b/graphrag_sdk/helpers.py
@@ -124,8 +124,11 @@ def validate_cypher_relations_exist(cypher: str, ontology: graphrag_sdk.Ontology
     not_found_relation_labels = []
     relation_labels = re.findall(r"\[:(.*?)\]", cypher)
     for label in relation_labels:
+        label = re.split(r"\d", label)[0].strip()
+        label = label.split(".")[0].strip() if "." in label else label
         label = label.split(":")[1] if ":" in label else label
         label = label.split("{")[0].strip() if "{" in label else label
+        label = label.split("*")[0].strip() if "*" in label else label
         if label not in [relation.label for relation in ontology.relations]:
             not_found_relation_labels.append(label)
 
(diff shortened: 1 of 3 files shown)


graphrag_sdk/helpers.py
```diff
@@ -124,8 +124,11 @@ def validate_cypher_relations_exist(cypher: str, ontology: graphrag_sdk.Ontology
     not_found_relation_labels = []
     relation_labels = re.findall(r"\[:(.*?)\]", cypher)
     for label in relation_labels:
+        label = re.split(r"\d", label)[0].strip()
+        label = label.split(".")[0].strip() if "." in label else label
         label = label.split(":")[1] if ":" in label else label
         label = label.split("{")[0].strip() if "{" in label else label
+        label = label.split("*")[0].strip() if "*" in label else label
         if label not in [relation.label for relation in ontology.relations]:
             not_found_relation_labels.append(label)
```
swilly22, line 133: That's not how I would go about this.
Start by getting the index of ':' and the index of ']'
Now what ever is in between call it `X`
See which of the followings has the minimum index: '*', '{' and ']'
the relationship is the string between the start index and the min index.

graphrag_sdk/helpers.py
```diff
@@ -144,7 +147,7 @@ def validate_cypher_relation_directions(
     for relation in relations:
         try:
             relation_label = (
-                re.search(r"(?:\[)(?:\w)*(?:\:)([^{\]]+)", relation.group(0))
+                re.search(r"(?:\[)(?:\w)*(?:\:)([^\d\{\]\*\.\:]+)", relation.group(0))
```
swilly22, line 150: I'll assume you've tested this.?

swilly22 (changes requested):

graphrag_sdk/helpers.py
galshubeli: yes

swilly22 (approved):

PR #48, opened 2024-12-18
Connection between Ontology graph & Knowledge graph 

galshubeli:


graphrag_sdk/kg.py
```diff
@@ -60,9 +61,17 @@ def __init__(
...
         self.graph = self.db.select_graph(name)
+        
+        # Load / Save ontology to database
+        if ontology is None:
+            ontology_graph = self.db.select_graph("{" + name + "}" + "_schema")
+            self._ontology = Ontology.from_graph(ontology_graph)
+        else:
+            ontology_graph = self.db.select_graph("{" + name + "}" + "_schema")
+            ontology.save_to_graph(ontology_graph)
+            self._ontology = ontology
```
swilly22, line 72:
```python
ontology_graph = self.db.select_graph("{" + name + "}" + "_schema")

if ontology is None:
            # load ontology from DB
            ontology = Ontology.from_graph(ontology_graph)
else:
            # save ontology to DB
            ontology.save_to_graph(ontology_graph)

if ontology is None:
            # probably throw an exception

self._ontology = ontology
```

swilly22 (changes requested):

graphrag_sdk/kg.py
galshubeli: fixed

swilly22 (approved):

PR #83, opened 2025-02-05
Fix Multiple Document Processing (e.g., PDFs) Handling

galshubeli:
- Fixed an issue with processing sources containing multiple documents, such as PDFs.
- Removed the sources property from the kg object.
- Modified the failed_sources property to failed_documents, which now contains the document IDs.

## Summary by CodeRabbit

- **New Features**
	- Documents now support an optional unique identifier, improving traceability by linking each document to its source.
	- A new check ensures that documents contain valid content.

- **Refactor**
	- Document creation has been standardized across all import methods for more consistent metadata handling.
	- The extraction and error tracking process has been streamlined for enhanced reliability.

### DIFF
--- a/graphrag_sdk/document_loaders/csv.py
+++ b/graphrag_sdk/document_loaders/csv.py
@@ -29,4 +29,4 @@ class CSVLoader:
                         ]
                     ]
                 )
-                yield Document(content)
+                yield Document(content, f"{self.path}#{i}")
--- a/graphrag_sdk/document.py
+++ b/graphrag_sdk/document.py
@@ -1,16 +1,24 @@
+from typing import Optional
+
+
 class Document():
     """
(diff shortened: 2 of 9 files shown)


graphrag_sdk/document_loaders/csv.py
```diff
@@ -29,4 +29,4 @@ def load(self) -> Iterator[Document]:
                         ]
                     ]
                 )
-                yield Document(content)
+                yield Document(content, f"{self.path}#{i}")
```
swilly22, line 32: this is a very expensive way to split a file into parts.
basically you want to count the number of lines in the file and break it according to `rows_per_document`

swilly22 (changes requested):

swilly22 (approved):

PR #202, opened 2026-04-01
Feat improve entity deduplication with hnswlib louvain detection and llm

Naseem77:


galshubeli:
**Review: networkx dependency & clustering approach**

The PR adds `networkx` as a dependency solely for `louvain_communities()` on a small in-memory similarity graph (~10-200 nodes per label group). A FalkorDB SDK depending on networkx for graph operations sends the wrong message.

More importantly, Louvain is designed for **graph topology** problems (social networks, citation graphs). Here the "graph" is just a **pairwise embedding similarity matrix with a threshold cutoff** — this is a metric space clustering problem, not a graph community detection problem.

The core architecture of the PR (3-tier: exact-match → hard merge → ambiguous zone with LLM verification) is solid. It's just the clustering method in the ambiguous zone that should change.

**Alternatives, ranked by fit:**

| Rank | Method | New dependency? | Memory | Fit for this use case | Production notes |
|---|---|---|---|---|---|
| 1 | **`scipy.cluster.hierarchy.fcluster`** (agglomerative) | No — scipy is already a transitive dep via numpy | O(N²) distance matrix, fine for N<1000 | **Best fit** — designed exactly for "cluster by pairwise distance with a threshold". Supports single/complete/average linkage to control chain risk. Deterministic, no seed needed. | Battle-tested, no build risk, zero added deps, trivial to swap in. Works directly on the cosine distance matrix already computed from hnswlib. |
| 2 | **`sklearn.cluster.DBSCAN`** | Likely already present (common transitive dep), otherwise ~30MB | O(N²) distance matrix or sparse input | **Great fit** — density-based, auto-detects cluster count, labels outliers explicitly (= "send to LLM"). Handles variable cluster sizes without specifying k. | Heavier install if not already present. `eps` parameter maps naturally to your similarity thresholds. Outlier labeling is a natural fit for the "ambiguous → LLM" flow. |
| 3 | **`scikit-network` Louvain** | Yes (~2.7-8.1 MB, C++/Cython) | Sparse matrix — memory efficient | **Decent fit** — same Louvain algorithm, drops networkx. But still a graph topology algorithm applied to a metric space problem. | C extension = build risk on Alpine/musl/ARM. New dep to pin and audit. Correct behavior, wrong abstraction. |
| 4 | **Union-Find (connected components)** | No — already in the codebase | O(N) — most memory efficient | **Partial fit** — simple and fast, but merges everything reachable via transitive chains. **All ambiguous pairs auto-merge, LLM verification never fires.** | Zero code, zero deps. But fundamentally changes the PR's design — the entire ambiguous→LLM flow becomes dead code. Risk of over-merging on larger label groups (200+ entities). |
| 5 | **`networkx` Louvain** (current) | Yes — `networkx>=3.0` (~2-3 MB, pure Python) | Full Graph object per label group — most wasteful for this use case | **Weakest fit** — correct output, but overkill general-purpose graph library for one function. Constructs a Graph object from similarity pairs just to feed them into clustering. Slowest option (pure Python). | A FalkorDB SDK shipping with networkx is an optics problem. Massive library surface for a single function call. |

**My recommendation**: Option 1 — `scipy.cluster.hierarchy.fcluster` with average or complete linkage. Zero new dependencies, semantically correct for pairwise distance clustering, deterministic, memory-efficient at our scale, and the distance matrix is already available from the hnswlib ANN step.

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/llm_verified_resolution.py
```diff
@@ -0,0 +1,520 @@
...
+        self,
+        graph_data: GraphData,
+        ctx: Context,
+    ) -> ResolutionResult:
+        ctx.log(
+            f"LLMVerifiedResolution: {len(graph_data.nodes)} nodes, "
+            f"{len(graph_data.relationships)} rels | "
+            f"hard={self.hard_threshold}, soft={self.soft_threshold}"
+        )
+
```
galshubeli, line 148: ~100 lines of Phase 1 logic here (exact-match grouping, description summarization, source_chunk_ids merging, property merge) is a near-copy of `SemanticResolution.resolve()`. This should be extracted into a shared helper or `LLMVerifiedResolution` should delegate Phase 1 to the existing `SemanticResolution` code. Duplicated logic means bugs get fixed in one place but not the other.

galshubeli (commented):
Code quality review — 8 issues found (see inline comments).

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/llm_verified_resolution.py
```diff
@@ -0,0 +1,520 @@
...
+                        end_node_id=end,
+                        type=rel.type,
+                        properties=rel.properties,
+                    )
+                )
+
+        ctx.log(
+            f"LLMVerifiedResolution complete: {len(deduplicated_nodes)} nodes "
+            f"({merged_count} merged), {len(deduplicated_rels)} rels"
+        )
```
galshubeli, line 185: `import hnswlib` inside the per-label loop body. This runs on every label iteration. Move to top-level with an optional-dependency guard:
```python
try:
    import hnswlib
except ImportError:
    hnswlib = None
```
Same applies to `import networkx as nx` below.

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/llm_verified_resolution.py
```diff
@@ -0,0 +1,520 @@
+# GraphRAG SDK 2.0 — Ingestion: LLM-Verified Resolution
```
galshubeli, line 1: No tests added for `LLMVerifiedResolution`. The repo has tests for `ExactMatchResolution`, `DescriptionMergeResolution`, and `SemanticResolution`. This new strategy needs coverage for at minimum: Phase 1 exact-match merging, hard-threshold embedding merges, ambiguous-zone LLM YES/NO handling, and relationship remapping/deduplication.

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/llm_verified_resolution.py
```diff
@@ -0,0 +1,520 @@
...
+            return "; ".join(f"{rel_type} -> {name}" for name, rel_type in nbrs)
+
+        # Group by label — never merge across labels
+        by_label: dict[str, list[GraphNode]] = defaultdict(list)
+        for n in nodes:
+            by_label[n.label].append(n)
+
+        remap: dict[str, str] = {}
+        total_hard = 0
+        total_llm = 0
```
galshubeli, line 233: Intra-community pairs get auto-merged without LLM verification. This is risky — if Louvain groups two genuinely different entities into the same community (because they are both connected to a shared ambiguous third entity), they merge silently. Consider sending a sample of intra-community pairs to LLM as a safety check, or at minimum logging which entities were auto-merged so accuracy regressions can be traced.

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/llm_verified_resolution.py
```diff
@@ -0,0 +1,520 @@
...
+
+        emb_cache: dict[str, list[float]] = ctx.metadata.setdefault("embedding_cache", {})
```
galshubeli, line 235: Using distinct defaults (`-1` vs `-2`) so unindexed nodes never match is clever but fragile. A future reader could easily "fix" this to use the same default and introduce a silent merge bug. Consider an explicit check instead:
```python
if gi in node_to_comm and gj in node_to_comm and node_to_comm[gi] == node_to_comm[gj]:
```

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/semantic_resolution.py
```diff
@@ -229,12 +237,47 @@ async def _fuzzy_merge(
...
+                        "SemanticResolution: '%s' node sample (up to 5 of %d): %s",
+                        label,
+                        len(miss_names),
+                        miss_names[:5],
+                    )
+                    miss_vecs = await _asyncio.wait_for(
+                        self.embedder.aembed_documents(miss_names),
+                        timeout=20.0,
+                    )
+                    for node, vec in zip(miss_nodes, miss_vecs):
```
galshubeli, line 162: `asyncio.wait_for(timeout=20)` around `aembed_documents` gives a false sense of cancellation. If the embedder uses `asyncio.to_thread` internally (which it does in `core/providers/base.py`), the timeout raises `TimeoutError` but the underlying thread keeps running and consuming resources. Either implement cancellation inside the embedder, or remove the `wait_for` and handle timeouts at the provider level.

graphrag_sdk/README.md
```diff
@@ -125,17 +125,28 @@ Override any pipeline step by passing a strategy:
...
+from graphrag_sdk.ingestion.resolution_strategies import (
+    ExactMatchResolution,
+    DescriptionMergeResolution,
+    SemanticResolution,
+    LLMVerifiedResolution,
 )
-
-# Use LLM for step 1 instead of GLiNER
+# Full 4-stage deduplication pipeline
+async def ingest_with_full_resolution(rag, path):
```
galshubeli, line 135: The strategy table and the "benchmark-winning pipeline" section have duplicated rows — both the old 2-strategy resolution content and the new 4-strategy content appear. Clean up the old entries.

graphrag_sdk/src/graphrag_sdk/ingestion/resolution_strategies/semantic_resolution.py
```diff
@@ -229,12 +237,47 @@ async def _fuzzy_merge(
...
             try:
-                vectors = await self.embedder.aembed_documents([str(name) for name in names])
-            except Exception:
-                logger.debug("Embedding failed for label %s, skipping", label, exc_info=True)
+                if miss_names:
+                    logger.info(
+                        "SemanticResolution: embedding %d '%s' nodes",
+                        len(miss_names),
+                        label,
+                    )
```
galshubeli, line 154: Full entity name list (`miss_names`) logged at INFO level. This is a PII risk and can produce massive log lines for large label groups. Move the name list to DEBUG and log only the count at INFO.

galshubeli (commented):
Code quality review — additional issues (see inline comments).

Naseem77:
@galshubeli 

1- replaced networkx/Louvain with scipy
Switched the clustering implementation to scipy.cluster.hierarchy.fcluster (average-linkage agglomerative). Removed the networkx dependency and added scipy>=1.11.

2- removed duplicated code:
moved exact_match_merge() and remap_relationships() from SemanticResolution and LLMVerifiedResolution into base.py, removing ~100 lines of duplicated.

3. fixed other comments.

galshubeli:
**Follow-up review: 4 remaining issues after fixes**

Good progress on the scipy swap, shared code extraction, and debug logging for auto-merges. A few things still open:

1. **`import hnswlib` still inside the loop in `semantic_resolution.py`** — `llm_verified_resolution.py` does it correctly at top-level with `try/except` (lines 11-14), but `semantic_resolution.py` still has `import hnswlib` inside `_fuzzy_merge` (runs on every label iteration). Should match the pattern used in `llm_verified_resolution.py`.

2. **`asyncio.wait_for` still present in `semantic_resolution.py`** — The `await _asyncio.wait_for(self.embedder.aembed_documents(...), timeout=20.0)` gives a false sense of cancellation. If the embedder uses `asyncio.to_thread` internally (which it does in `core/providers/base.py`), the timeout raises `TimeoutError` but the thread keeps running. Either remove it or handle timeouts at the provider level.

3. **README still references "Louvain community detection"** — The benchmark-winning pipeline description says "Louvain community detection + batched LLM verification" but the implementation now uses scipy agglomerative clustering. Should be updated to match.

4. **No tests for `LLMVerifiedResolution`** — The repo has tests for all other resolution strategies. This new strategy still has zero test coverage.

galshubeli (approved):

PR #213, opened 2026-04-15
Docs/benchmark guide

Naseem77:


docs/benchmark.md
```diff
@@ -1,451 +1,421 @@
...
-Criteria: factual correctness, completeness, and relevance to the question.
+| Phase | Duration |
+|-------|--------:|
+| Avg. query latency | 3.6 s |
 
-### Variance
+### LLM-as-judge scores (supplementary)
 
-Approximately 52% of questions show score variance between runs due to LLM generation non-determinism (even with temperature=0). The ~85% accuracy represents a practical ceiling for single-pass RAG with GPT-4.1 on this corpus.
+Independently scored by GPT-4o on a 100-question sample:
```
galshubeli: ?

Naseem77: updated.

galshubeli (approved):

galshubeli:
Review punch list — remaining items in `docs/benchmark.md` to address before merge (the README caption is already aligned in `cc06664`):

1. **`temperature=0.0` was removed** (previously at L402). Without it, LLM-judged scores drift between runs — the old doc explicitly set it. One-line fix in the Step 1 Azure LLM config.

2. **Leaderboard table has no "these are published numbers" caveat** (around L734–739). As written, a reader assumes all 9 systems were run under identical conditions by us. Add a sentence near the table stating that only the FalkorDB row was run by us; competitor numbers come from the GraphRAG-Bench published leaderboard (with a snapshot date, since leaderboards change).

3. **`ChainedResolution` example is broken** (around L346). The loop does `graph_data = await stage.resolve(graph_data, ctx)` — but `ResolutionStrategy.resolve()` returns a `ResolutionResult`, not `GraphData`. Iteration 2+ passes the wrong type. Either unwrap `.graph_data` between stages or drop the example.

4. **Env var rename: `AZURE_OPENAI_API_VERSION` → `AZURE_API_VERSION`** (L78). `.env.example` and other places in the repo use `AZURE_OPENAI_API_VERSION`. Intentional? LiteLLM accepts both, but consistency is easier on users.

#3 is the only hard breakage (copy-paste example that won't execute). #1 and #2 are credibility concerns for a public-facing benchmark doc. #4 is a nit.

PR #255, opened 2026-05-14
feat(retrieval): CypherFirstAggregationStrategy + cypher-path accuracy fixes

galshubeli:
## Summary

Adds a new opt-in retrieval strategy, `CypherFirstAggregationStrategy`, that routes quantitative/structural questions ("how many", "which X has the most", "BOTH A and B", "are there any X without Y", "what is the average …") through a deterministic Cypher-first path. Non-aggregation questions delegate to `MultiPathRetrieval` unchanged.

Alongside the strategy, this PR also lands six targeted SDK-level fixes to the pre-existing cypher path (used by `MultiPathRetrieval(enable_cypher=True)` too): smart `LIMIT` injection, `APOC`/`GDS`/`db.*` function-call blocklist in the validator, 0-row label-widen fallback, two new schema-prompt examples, and metadata threading for observability.

## Background

A focused failure-mode investigation on a 56-person / 10-org synthetic benchmark identified seven distinct failure modes on aggregation questions — none of them random one-offs:

- Cypher results encoded as `\" | \"`-joined strings lose column names, so the answer-LLM can swap row values
- Auto `LIMIT 25` truncates group-by aggregations silently
- `apoc.text.regexGroups(...)`-style function calls slipped past the `CALL` blocklist
- Typed-label cypher queries fail when extraction labeled the entity differently
- Schema prompt had no examples for "BOTH X AND Y" or top-N group-by
- Empty cypher results were treated identically for positive and negation existentials
- Roles/projects extracted as free text in chunks weren't queryable as typed nodes

The new strategy directly addresses each one. The 7-question failure-mode benchmark went from **2-5/7** (depending on which pre-fix variant) to **7/7 stable** across three runs.

## What's in this PR (4 commits)

| Commit | Title |
|---|---|
| \`bb920c6\` | improve text-to-Cypher accuracy on aggregation questions |
| \`7fa7a1f\` | add CypherFirstAggregationStrategy |
| \`deaa64d\` | pre-merge review fixes for CypherFirstAggregationStrategy |
| \`d818eb1\` | split CypherFirst into path classes + pluggable extractor |

Each commit has its own detailed body.

## Compatibility (cypher-off callers)

The PR is **additive** for users who don't opt in:
- `CypherFirstAggregationStrategy` only runs if passed explicitly as `retrieval_strategy=`
- The cypher-path improvements (smart LIMIT, APOC blocklist, etc.) live inside the cypher generation/execution code and only run when `enable_cypher=True`
- The "Authoritative Graph Query Results" system-prompt rule is added conditionally — only when the retriever produced a cypher_results section. With cypher off, the base system prompt is **byte-identical** to main.

Test coverage (\`tests/test_facade.py::TestCypherAuthorityRuleInjection\`) pins the contract.

## Usage

\`\`\`python
from graphrag_sdk import (
    CypherFirstAggregationStrategy,
    FastCorefResolver,
    GraphExtraction,
    LLMVerifiedResolution,
    SentenceTokenCapChunking,
)

chunker = SentenceTokenCapChunking(max_tokens=512, overlap_sentences=2)
extractor = GraphExtraction(llm=llm, coref_resolver=FastCorefResolver())
resolver = LLMVerifiedResolution(llm=llm, embedder=embedder)

async with GraphRAG(
    connection=conn, llm=llm, embedder=embedder, embedding_dimension=256,
) as rag:
    await rag.ingest(text=doc, chunker=chunker,
                     extractor=extractor, resolver=resolver)
    await rag.finalize()
    rag._retrieval_strategy = CypherFirstAggregationStrategy(
        graph_store=rag._graph_store,
        vector_store=rag._vector_store,
        embedder=embedder,
        llm=llm,
    )
    answer = await rag.completion("Which city has the most employees?")
\`\`\`

## Related PRs

- **#254** — independent change to make `SentenceTokenCapChunking` the default chunker (recommended pairing; the benchmark required it to hit 7/7)
- **#253** — currently mis-targeted, will be retargeted to stack on top of this PR so its diff shrinks to the 33-line docstring follow-up

## Test plan

- [x] Unit tests: \`tests/test_cypher_first.py\` (66 tests), \`tests/test_cypher_generation.py\` (49 tests), \`tests/test_facade.py\` (75 tests including R1 contract)
- [x] End-to-end aggregation benchmark: 7/7 stable across 3 runs (\`CypherFirstAggregationStrategy\`) and 7/7 single run (\`MultiPathRetrieval(enable_cypher=True)\`)
- [x] Full SDK suite: 748 passed, 24 skipped on the cypher branch tip
- [x] \`ruff check\` clean on all touched files
- [ ] Reviewer to confirm CI green on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Cypher-first retrieval for aggregation/numeric queries
  * Pluggable phrase-extraction interface with a default extractor
  * Exposed new aggregation and phrase-extractor classes in the public API
  * Markdown-table formatting and numeric-coercion utilities

* **Improvements**
  * System prompt addendum when authoritative graph-query results are present
  * Stricter query validation, default LIMIT injection, and label‑widening retry on empty results
  * Increased graph-result row cap and richer cypher-related metadata

* **Tests**
  * Extensive tests covering routing, generation, formatting, coercion, and authority-rule behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/FalkorDB/GraphRAG-SDK/pull/255?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_first.py
+++ b/graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_first.py
@@ -42,9 +42,7 @@ import statistics
 from typing import Any
 
 from graphrag_sdk.core.context import Context
-from graphrag_sdk.core.exceptions import RetrieverError
 from graphrag_sdk.core.models import (
-    ChatMessage,
     RawSearchResult,
     RetrieverResult,
     RetrieverResultItem,
@@ -364,6 +362,31 @@ def _wrap_authoritative(body: str, *, source_note: str = "") -> str:
     return f"{_AUTH_HEADING}{note}\n\n{body}"
 
 
+# Canonical labels for the sub-path metadata key ``cypher_first_path``.
+# Operators / metrics dashboards can group on these to see which path
+# fires for each query.
+PATH_NUMERIC_MATH = "numeric_math"
+PATH_SHARED_PROPERTY_HYBRID = "shared_property_hybrid"
+PATH_CYPHER_TABLE = "cypher_table"
+PATH_NEGATION_EMPTY_NO = "negation_empty_no"
+PATH_RAG_FALLBACK = "rag_fallback"
+PATH_RAG_FALLBACK_NUMERIC_FAIL = "rag_fallback_numeric_fail"
+PATH_RAG_FALLBACK_CYPHER_EMPTY = "rag_fallback_cypher_empty"
+
+
+def _tag_path(result: RawSearchResult, path: str) -> RawSearchResult:
+    """Attach the ``cypher_first_path`` label to a strategy result.
+
+    Used both for results we construct ourselves and for results returned
+    from the delegated RAG fallback strategy — operators get a uniform
+    signal regardless of which branch handled the query.
+    """
+    meta = dict(result.metadata or {})
+    meta.setdefault("strategy", "cypher_first")
+    meta["cypher_first_path"] = path
+    return RawSearchResult(records=result.records, metadata=meta)
+
+
 # ─────────────────────────────────────────────────────────────────
 # Strategy
 # ─────────────────────────────────────────────────────────────────
@@ -383,6 +406,46 @@ class CypherFirstAggregationStrategy(RetrievalStrategy):
 
     Safe as the top-level strategy on ``GraphRAG``: non-aggregation
     questions get the existing pipeline unchanged.
+
+    Every returned :class:`RawSearchResult` carries a ``cypher_first_path``
+    metadata key whose value is one of ``PATH_*`` module constants — useful
+    for operator dashboards that want to see which sub-path fired.
+
+    Assumptions and known limits
+    ----------------------------
+    The shared-property hybrid (M5) was tuned on graphs produced by the
+    SDK's default ``GraphExtraction`` pipeline. It makes the following
+    assumptions; when they're violated, the hybrid silently returns
+    ``None`` and the strategy falls back to the multi-candidate Cypher
+    path (which still works, just without free-text recovery):
+
+    - **Graph topology.** Organizations are connected to Persons via
+      ``[:RELATES]``, and Persons are connected to ``Chunk`` nodes via
+      ``[:MENTIONED_IN]``. This is the canonical shape the SDK builds.
+      Custom extractors that use different edge types or skip chunk
+      provenance will not benefit from M5 — the strategy logs a warning
+      and continues.
+    - **Prose shape.** Role and project values are extracted by regex
+      from Person descriptions and chunk text. The regexes target the
+      phrasing patterns ``"works at X as a <role>"`` and ``"contributes
+      to <project>"``. Domains whose prose departs from these patterns
+      (medical / legal / e-commerce / non-English) will not match — the
+      result is empty role/project sets, not wrong answers, but the
+      hybrid won't help.
+    - **Role vocabulary.** The role extractor anchors on the suffixes
+      ``engineer | scientist | manager | architect | researcher |
+      developer | analyst | specialist | designer | consultant``. Other
+      job titles ("director", "VP", "lead") won't match.
+
+    Accuracy ceiling
+    ----------------
+    The strategy faithfully returns what is in the graph. Duplicate
+    entities ("Wayne En" vs "Wayne Enterprises"), chunk-boundary-truncated
+    names, or non-deduplicated short-form references ("Carla" vs "Carla
+    Okafor") all flow through into the answer. Cypher counts will be
+    inflated; "which X" lists will contain duplicates. These are
+    extraction-quality issues — not strategy bugs — and should be
+    addressed in the ingestion pipeline (resolver, coref, dedup).
 
     Args:
         graph_store: Required for Cypher execution.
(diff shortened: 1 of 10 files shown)


Naseem77 (changes requested):
1. [medium] graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_generation.py:222 Dotted-namespace blocklist matches inside string literals
 - Problem: validate_cypher only strips comments (//, /*
  */) from cypher_norm — it doesn't strip string literals before running _DOTTED_FN_RE. A legitimate query like MATCH (n) WHERE n.text CONTAINS 'apoc.foo(' RETURN n would be rejected as using an APOC
function.
 - Impact: False-positive rejection forces the retry loop to regenerate (or, after exhausting retries, the strategy silently returns empty results) for queries whose string-literal payloads happen to contain
<word>.<word>(. The pattern is realistic for LLM-generated regex/text-search predicates.
 - Fix: Strip single/double-quoted string literals (with escape handling) from cypher_norm before applying _DOTTED_FN_RE, the same way comments are stripped. The existing \bCALL\b check has the same latent
issue but the new dotted-fn check is much more likely to hit it because dots inside quoted regexes are common.
 2. [medium] graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_first.py:580 Shared-property hybrid runs an unbounded full-graph scan
 - Problem: The batch_cypher MATCH (o:Organization)<-[:RELATES]-(p:Person) OPTIONAL MATCH (p)-[:MENTIONED_IN]->(c:Chunk) RETURN ... collect(DISTINCT c.text) AS chunks has no LIMIT and pulls every Person–Org
pair with all their chunk texts into memory per query. _sanitize_cypher is not applied (the query goes straight to graph_store.query_raw).
 - Impact: On any non-toy graph (thousands of persons × dozens of chunks each), this is a multi-MB result on every "both A and B" / "same role as" question. Latency and memory grow linearly with graph size,
with no cap.
 - Fix: Restrict the batch to the orgs named in the question (the regex already captures org_a/org_b / target before the call) and/or cap chunk text length and row count. The current code already discards
rows whose org name doesn't match — push that filter into the query.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_generation.py
```diff
@@ -206,22 +312,39 @@ def validate_cypher(cypher: str) -> list[str]:
...
+    if re.search(r"\bCALL\b", cypher_code, re.IGNORECASE):
         errors.append("CALL procedures are not allowed in generated queries")
-    if re.search(r"\bLOAD\s+CSV\b", cypher_norm, re.IGNORECASE):
+    if re.search(r"\bLOAD\s+CSV\b", cypher_code, re.IGNORECASE):
         errors.append("LOAD CSV is not allowed in generated queries")
 
+    # Reject dotted-namespace function calls (apoc.*, gds.*, db.*).
+    # FalkorDB doesn't implement these plugins; the call silently returns 0
+    # rows at execution. Surfacing it here lets the retry loop regenerate.
+    for ns, _ in _DOTTED_FN_RE.findall(cypher_code):
```
galshubeli, line 349: @Naseem77 — good catch on the literal-stripping gap. Fixed in 4c4f54c.

The validator now strips both single- and double-quoted Cypher string literals (with `\`-escape handling so `'apoc\'s foo('` doesn't terminate the literal early) before running the `_DOTTED_FN_RE`, `\bCALL\b`, `LOAD CSV`, and write-keyword checks. Structural checks that can't legitimately occur inside a string predicate (start-keyword allowlist, `RETURN` presence, label existence) still run on `cypher_norm` since they don't have the false-positive risk.

Regression test in `test_cypher_generation.py::test_does_not_reject_blocklisted_tokens_inside_string_literals` covers:
- `n.text CONTAINS 'apoc.foo('`
- `n.text CONTAINS "gds.shortest.path("`
- `n.title = 'CALL me maybe'`
- escape-honoring: `n.text = 'apoc\'s db.foo('`
- `n.text CONTAINS 'LOAD CSV here'`

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_first.py
```diff
@@ -0,0 +1,1129 @@
...
+            if not org_a or not org_b:
+                return None
+            where_clause = (
+                "WHERE toLower(o.name) CONTAINS toLower($org_a) "
+                "OR toLower(o.name) CONTAINS toLower($org_b) "
+            )
+            params["org_a"] = org_a
+            params["org_b"] = org_b
+
+        batch_cypher = (
```
galshubeli, line 622: @Naseem77 — agreed, the unbounded scan needed bounds. Fixed in 652f2a4.

Three changes to the batch:
1. **For "both A and B"**: push `WHERE toLower(o.name) CONTAINS toLower($org_a) OR toLower(o.name) CONTAINS toLower($org_b)` into the query. `_BOTH_AB_RE` already captures both org names, so this just propagates them into Cypher rather than re-deriving them downstream in `_gather`.
2. **For "same X as Z"**: no org filter is applicable (we need every other org to compare against the target), so the LIMIT + substring cap below are the only guardrails for this shape.
3. **For both shapes**: `collect(DISTINCT substring(c.text, 0, $max_chunk_len))` (cap = 2000 chars — phrase extraction runs on sentence fragments anyway, so the head of each chunk carries the signal) and a `LIMIT 500` row cap as a safety net.

The two caps live as module-level `_HYBRID_BATCH_ROW_CAP` / `_HYBRID_CHUNK_TEXT_CAP` constants with rationale in their docstring. Regression test `test_hybrid_batch_query_is_scoped_and_capped` asserts the WHERE / LIMIT / substring / params behavior for both shapes.

I didn't run `_sanitize_cypher` on the batch — it's an internal template, not LLM-generated, so it doesn't need the LLM-pattern cleanup. Happy to add it as belt-and-suspenders if you'd prefer.

PR #261, opened 2026-05-18
Reliability hardening suite

### LINKED ISSUE #257: Production Hardening & Release Readiness
- [x]   - 0.1 LLM & Embedding Timeout Enforcement — asyncio.wait_for + typed timeout errors
- [x]   - 0.2 Error Visibility — audit 73 except Exception blocks, add logger.error(), typed errors
- [x]   - 0.3 Async Context Manager — async with GraphRAG() support, connection cleanup
- [x]   - 0.4 Latency Budget Enforcement — check budget before each retrieval phase & LLM call
- [x]   - 0.5 Integration Tests with Real FalkorDB — docker-compose + CI integration test job
- [x]   - 0.6 Release Automation — PyPI publish workflow, docs deployment, dependabot

All task list waiting for review in #261

Naseem77:
## Summary
- enforce LLM and embedding timeouts with typed timeout errors
- improve error visibility for broad exception handling paths
- cover GraphRAG async context manager cleanup behavior
- enforce latency budgets across retrieval and LLM phases
- mark and wire real FalkorDB integration tests
- tighten release automation for docs, PyPI checks, artifacts, and Dependabot

## Review
- ran the local review process after each task and only committed approved fixes

## Testing
- targeted provider, facade, retrieval, integration-marker, YAML, and build checks were run during each task

## Summary by CodeRabbit

* **New Features**
  * Timeouts for LLM/embedding calls, a new LatencyBudgetExceededError, and latency-budget enforcement across retrieval and completion.

* **Bug Fixes**
  * Improved error logging and clearer, typed timeout/error propagation so failures are reported consistently.

* **Documentation**
  * Dev setup now uses Docker Compose; added instructions to run integration tests locally.

* **Chores**
  * CI/CD workflow tweaks and weekly Dependabot updates.

* **Tests**
  * Many new tests for timeouts, budget checkpoints, and propagation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FalkorDB/GraphRAG-SDK/pull/261?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
+  - package-ecosystem: "pip"
+    directory: "/graphrag_sdk"
+    schedule:
+      interval: "weekly"
+    groups:
+      python-dependencies:
+        patterns:
+          - "*"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,7 @@ services:
     image: falkordb/falkordb:v4.18.0
     ports:
       - "6379:6379"
+      - "3000:3000"
     volumes:
       - falkordb_data:/data
     healthcheck:
@@ -10,6 +11,7 @@ services:
       interval: 5s
       timeout: 3s
       retries: 5
+      start_period: 5s
 
 volumes:
   falkordb_data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     image: falkordb/falkordb:v4.18.0
     ports:
       - "6379:6379"
-      - "3000:3000"
+      - "127.0.0.1:3000:3000"
     volumes:
       - falkordb_data:/data
     healthcheck:
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -1480,13 +1480,16 @@ class GraphRAG:
                         f"history[{i}]: each message must have 'role' and "
                         f"'content' keys, got {sorted(msg.keys())}"
                     )
-                try:
-                    validated.append(ChatMessage(role=msg["role"], content=msg["content"]))
-                except Exception:
+                role = msg["role"]
+                content = msg["content"]
+                if role not in {"system", "user", "assistant"}:
                     raise ValueError(
-                        f"history[{i}]: invalid role '{msg['role']}'. "
+                        f"history[{i}]: invalid role '{role}'. "
                         f"Must be one of: 'system', 'user', 'assistant'"
                     )
+                if not isinstance(content, str):
+                    raise ValueError(f"history[{i}]: content must be a string")
+                validated.append(ChatMessage(role=role, content=content))
             else:
                 raise TypeError(
                     f"history[{i}]: expected ChatMessage or dict, got {type(msg).__name__}"
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -1527,7 +1527,7 @@ class GraphRAG:
         )
         try:
             ctx.ensure_budget("question rewrite LLM call")
-            resp = await self.llm.ainvoke(prompt)
+            resp = await self.llm.ainvoke(prompt, timeout=ctx.remaining_budget_seconds)
             rewritten = (resp.content or "").strip().splitlines()[0].strip() if resp.content else ""
         except LatencyBudgetExceededError:
             raise
@@ -1652,7 +1652,10 @@ class GraphRAG:
         ]
 
         ctx.ensure_budget("completion LLM call")
-        llm_response = await self.llm.ainvoke_messages(messages)
+        llm_response = await self.llm.ainvoke_messages(
+            messages,
+            timeout=ctx.remaining_budget_seconds,
+        )
 
         result = RagResult(
             answer=self._clean_answer(llm_response.content),
@@ -1692,7 +1695,7 @@ class GraphRAG:
         except Exception:
             logger.debug("Failed to write graph config node", exc_info=True)
 
-    async def _validate_graph_config(self, ctx: Context | None = None) -> None:
+    async def _validate_graph_config(self, *, ctx: Context | None = None) -> None:
         """Check that the current embedder matches the graph's stored config.
 
         Two checks, both cached after first run:
@@ -1753,7 +1756,10 @@ class GraphRAG:
         if ctx is not None:
             ctx.ensure_budget("graph config embedder probe")
         try:
-            probe = await self.embedder.aembed_query("dim_check")
+            probe = await self.embedder.aembed_query(
+                "dim_check",
+                timeout=ctx.remaining_budget_seconds if ctx is not None else None,
+            )
         except LatencyBudgetExceededError:
             raise
         except Exception:
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -1527,7 +1527,10 @@ class GraphRAG:
         )
         try:
             ctx.ensure_budget("question rewrite LLM call")
-            resp = await self.llm.ainvoke(prompt, timeout=ctx.remaining_budget_seconds)
+            resp = await self.llm.ainvoke(
+                prompt,
+                timeout=ctx.provider_timeout_seconds("question rewrite LLM call"),
+            )
             rewritten = (resp.content or "").strip().splitlines()[0].strip() if resp.content else ""
         except LatencyBudgetExceededError:
             raise
@@ -1654,7 +1657,7 @@ class GraphRAG:
         ctx.ensure_budget("completion LLM call")
         llm_response = await self.llm.ainvoke_messages(
             messages,
-            timeout=ctx.remaining_budget_seconds,
+            timeout=ctx.provider_timeout_seconds("completion LLM call"),
         )
 
         result = RagResult(
@@ -1758,7 +1761,11 @@ class GraphRAG:
         try:
             probe = await self.embedder.aembed_query(
                 "dim_check",
-                timeout=ctx.remaining_budget_seconds if ctx is not None else None,
+                timeout=(
+                    ctx.provider_timeout_seconds("graph config embedder probe")
+                    if ctx is not None
+                    else None
+                ),
             )
         except LatencyBudgetExceededError:
             raise
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -15,7 +15,12 @@ from uuid import uuid4
 from graphrag_sdk import __version__
 from graphrag_sdk.core.connection import ConnectionConfig, FalkorDBConnection
 from graphrag_sdk.core.context import Context
-from graphrag_sdk.core.exceptions import ConfigError, DatabaseError, DocumentNotFoundError
+from graphrag_sdk.core.exceptions import (
+    ConfigError,
+    DatabaseError,
+    DocumentNotFoundError,
+    LatencyBudgetExceededError,
+)
 from graphrag_sdk.core.models import (
     ApplyChangesResult,
     BatchEntry,
@@ -1446,13 +1451,16 @@ class GraphRAG:
             ctx = Context()
 
         ctx.log(f"Retrieve: {question[:80]}...")
+        ctx.ensure_budget("graph config validation")
 
-        await self._validate_graph_config()
+        await self._validate_graph_config(ctx=ctx)
 
         retrieval = strategy or self._retrieval_strategy
+        ctx.ensure_budget("retrieval strategy search")
         retriever_result = await retrieval.search(question, ctx)
 
         if reranker is not None:
+            ctx.ensure_budget("retrieval reranking")
             retriever_result = await reranker.rerank(question, retriever_result, ctx)
 
         ctx.log(f"Retrieved {len(retriever_result.items)} context items")
@@ -1518,8 +1526,11 @@ class GraphRAG:
             question=question,
         )
         try:
+            ctx.ensure_budget("question rewrite LLM call")
             resp = await self.llm.ainvoke(prompt)
             rewritten = (resp.content or "").strip().splitlines()[0].strip() if resp.content else ""
+        except LatencyBudgetExceededError:
+            raise
         except Exception as e:
             # Broad catch is intentional (see docstring) — but log at WARNING
             # with full traceback so programming bugs surface in operator
@@ -1588,6 +1599,7 @@ class GraphRAG:
         # Step 1: Optionally rewrite the question for retrieval.
         retrieval_query = question
         if validated_history and rewrite_question_with_history:
+            ctx.ensure_budget("question rewrite")
             retrieval_query = await self._rewrite_question_with_history(
                 question,
                 validated_history,
@@ -1597,6 +1609,7 @@ class GraphRAG:
                 ctx.log(f"Rewrote for retrieval: {retrieval_query[:80]}")
 
         # Step 2: Retrieve + rerank (using possibly-rewritten query).
+        ctx.ensure_budget("completion retrieval")
         retriever_result = await self.retrieve(
             retrieval_query,
             strategy=strategy,
@@ -1638,6 +1651,7 @@ class GraphRAG:
             ChatMessage(role="user", content=final_user_content),
         ]
 
+        ctx.ensure_budget("completion LLM call")
         llm_response = await self.llm.ainvoke_messages(messages)
 
         result = RagResult(
@@ -1678,7 +1692,7 @@ class GraphRAG:
         except Exception:
             logger.debug("Failed to write graph config node", exc_info=True)
 
-    async def _validate_graph_config(self) -> None:
+    async def _validate_graph_config(self, ctx: Context | None = None) -> None:
         """Check that the current embedder matches the graph's stored config.
 
         Two checks, both cached after first run:
@@ -1700,6 +1714,8 @@ class GraphRAG:
             return
 
         try:
+            if ctx is not None:
+                ctx.ensure_budget("graph config query")
             result = await self._graph_store.query_raw(
                 "MATCH (c:__GraphRAGConfig__ {id: 'default'}) "
                 "RETURN c.embedding_model, c.embedding_dimension"
@@ -1724,6 +1740,8 @@ class GraphRAG:
                     )
         except ConfigError:
             raise
+        except LatencyBudgetExceededError:
+            raise
         except Exception:
             # Don't mark as validated on transient failures — retry next call.
             logger.debug("Failed to validate graph config", exc_info=True)
@@ -1732,8 +1750,12 @@ class GraphRAG:
         # Probe the embedder once: confirm it produces vectors of the
         # configured dimension. Catches user error like
         # ``embedding_dimension=256`` paired with a 1536-dim model.
+        if ctx is not None:
+            ctx.ensure_budget("graph config embedder probe")
         try:
             probe = await self.embedder.aembed_query("dim_check")
+        except LatencyBudgetExceededError:
+            raise
         except Exception:
             # Probe failure is non-fatal — but don't cache a "validated"
             # state, otherwise a transient outage permanently disables
--- a/graphrag_sdk/src/graphrag_sdk/core/connection.py
+++ b/graphrag_sdk/src/graphrag_sdk/core/connection.py
@@ -210,7 +210,6 @@ class FalkorDBConnection:
                 "FalkorDB query failure details",
                 exc_info=(type(last_exc), last_exc, last_exc.__traceback__),
             )
-        if last_exc is not None:
             raise last_exc
         raise RuntimeError("FalkorDB query failed without an exception")
 
--- a/graphrag_sdk/src/graphrag_sdk/core/connection.py
+++ b/graphrag_sdk/src/graphrag_sdk/core/connection.py
@@ -11,6 +11,8 @@ from dataclasses import dataclass, field
 from typing import Any
 from urllib.parse import urlparse
 
+from graphrag_sdk.core.exceptions import DatabaseError
+
 logger = logging.getLogger(__name__)
 
 
@@ -180,7 +182,13 @@ class FalkorDBConnection:
                 last_exc = exc
                 # Don't retry non-transient errors (e.g. schema/index conflicts)
                 if self._is_non_transient(exc):
-                    raise
+                    logger.error(
+                        "Non-transient FalkorDB query failure: %s: %s",
+                        type(exc).__name__,
+                        exc,
+                    )
+                    logger.debug("Non-transient FalkorDB query failure details", exc_info=True)
+                    raise DatabaseError(f"FalkorDB query failed: {exc}") from exc
                 await self._breaker.record_failure()
                 logger.warning(
                     "Query attempt %d/%d failed: %s",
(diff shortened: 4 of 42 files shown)


graphrag_sdk/src/graphrag_sdk/core/connection.py
```diff
@@ -180,7 +182,13 @@ async def query(
...
                 # Don't retry non-transient errors (e.g. schema/index conflicts)
                 if self._is_non_transient(exc):
-                    raise
+                    logger.error(
+                        "Non-transient FalkorDB query failure: %s: %s",
+                        type(exc).__name__,
+                        exc,
+                    )
+                    logger.debug("Non-transient FalkorDB query failure details", exc_info=True)
+                    raise DatabaseError(f"FalkorDB query failed: {exc}") from exc
```
galshubeli: 🔴 **C1 — Breaking change to a public class, undocumented.**

Previously the original exception (`falkordb.ResponseError`, `redis.ConnectionError`, etc.) propagated untouched. Now every non-transient error is wrapped in `DatabaseError`. Downstream code like:

```python
try:
    await conn.query(...)
except falkordb.ResponseError as e:    # ← silently stops matching
    if 'index' in str(e): ...
```

…breaks. `FalkorDBConnection` is re-exported from `graphrag_sdk/__init__.py:21`, so this is a public-class contract change.

Either (a) preserve the original type (just `raise` it untouched and emit the structured ERROR log alongside), or (b) advertise as breaking — major bump + CHANGELOG migration note. As written it's an opaque API change buried in a 1300-line PR.

galshubeli (commented):
Deep-dive review of the reliability hardening suite. Solid intent and real value, but three load-bearing issues need addressing before merge. Posted as line comments — summary below.

**Blocking (Critical):**
- **C1**: `connection.py` silently wraps every DB error in `DatabaseError` — breaking change for any downstream `except falkordb.<Specific>Error` clause, and `FalkorDBConnection` is a re-exported public class.
- **C2**: the latency budget is never wired into provider `timeout=`. Once a slow LLM call is in flight, the budget can fly past the threshold and nothing aborts it.
- **C3**: `aembed_documents(..., timeout=10)` applies the timeout *per batch* (and per binary-split sub-call), not as an overall deadline — the public contract reads as 10s but wall-clock total can be 50s+.

**Wins worth keeping:** the MENTIONED_IN cosine-rank merge from #259, the history-validation refactor in `api/main.py` (error messages were wrong before, are right now), the secret-leakage test in `test_providers.py`, the PdfLoader open/finally fix. Tests pass cleanly (393/393 on touched files).

graphrag_sdk/src/graphrag_sdk/core/connection.py
```diff
@@ -193,7 +201,18 @@ async def query(
...
+            self.config.retry_count,
+            type(last_exc).__name__ if last_exc is not None else "UnknownError",
+            last_exc,
+        )
+        if last_exc is not None:
+            logger.debug(
+                "FalkorDB query failure details",
+                exc_info=(type(last_exc), last_exc, last_exc.__traceback__),
+            )
+        raise DatabaseError(f"FalkorDB query failed: {last_exc}") from last_exc
```
galshubeli: 🔴 **C1 (same issue) — retries-exhausted path.**

`raise last_exc` → `raise DatabaseError(...) from last_exc` is the same contract change in the transient-retry-exhausted path. Same fix applies: re-raise the original.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/multi_path.py
```diff
@@ -186,33 +187,40 @@ async def _execute(
...
     ) -> RawSearchResult:
         # 1. Extract keywords
-        simple_kw, llm_kw = await self._extract_keywords(query)
+        ctx.ensure_budget("MultiPath keyword extraction")
+        simple_kw, llm_kw = await self._extract_keywords(query, ctx)
         all_keywords = llm_kw[:8] + simple_kw
         ctx.log(f"MultiPath [1/9]: {len(all_keywords)} keywords extracted")
 
         # 2. Embed question only
+        ctx.ensure_budget("MultiPath question embedding")
```
galshubeli, line 204: 🔴 **C2 — Budget never reaches the network call.**

This (and ~25 other `ctx.ensure_budget(...)` sites) only check at phase boundaries. Once `await self._embedder.aembed_query(query)` is in flight with `timeout=None`, an unresponsive provider can hang for minutes — the budget check on the next phase only notices *after* the call returns.

This is the load-bearing missing wire of the whole PR. One-liner fix at each provider call site:

```python
remaining_s = (ctx.remaining_budget_ms or 0) / 1000.0 or None
query_vector = await self._embedder.aembed_query(query, timeout=remaining_s)
```

The `timeout=` plumbing is already in place on every provider method — just connect the two halves. Without this, the PR enforces a 'phase-boundary budget,' which is useful but not what the description claims.

graphrag_sdk/src/graphrag_sdk/core/providers/litellm.py
```diff
@@ -307,16 +363,42 @@ def embed_documents(self, texts: list[str], **kwargs: Any) -> list[list[float]]:
...
+    ) -> list[list[float]]:
+        _validate_timeout(timeout)
         if not texts:
             return []
         results: list[list[float]] = []
         for start in range(0, len(texts), self.batch_size):
             batch = texts[start : start + self.batch_size]
-            results.extend(await binary_split_retry_async(self._raw_embed_async, batch, **kwargs))
+            results.extend(
+                await binary_split_retry_async(
```
galshubeli, line 415: 🔴 **C3 — `timeout` is per-batch, not a deadline.**

With `aembed_documents(texts=[100], timeout=10)` and `batch_size=20`, the public API reads as 10s, but wall-clock total can reach 50s+ (5 batches × 10s), plus extra sub-calls from `binary_split_retry_async`. The new test `test_aembed_documents_timeout_is_not_binary_split` only proves 'no retry on timeout' on a single batch.

Options:
```python
# Option A: rename to clarify
async def aembed_documents(self, texts, *, per_batch_timeout=None, ...):

# Option B: implement an actual deadline
deadline = time.monotonic() + timeout if timeout else None
for start in range(0, len(texts), self.batch_size):
    remaining = deadline - time.monotonic() if deadline else None
    if remaining is not None and remaining <= 0:
        raise EmbeddingTimeoutError(...)
    await binary_split_retry_async(..., timeout=remaining)
```

graphrag_sdk/src/graphrag_sdk/core/providers/base.py
```diff
@@ -13,12 +13,35 @@
...
 from pydantic import BaseModel
 
+from graphrag_sdk.core.exceptions import EmbeddingTimeoutError, LLMTimeoutError
 from graphrag_sdk.core.models import ChatMessage, LLMResponse
 from graphrag_sdk.core.providers._retry import summarize_exception
 
 logger = logging.getLogger(__name__)
 
 
+async def _wait_for_provider_call(
```
galshubeli: 🟠 **H1 — `_`-prefixed helpers imported across modules.**

Leading underscore = module-private in Python. But `litellm.py:21` and `openrouter.py:21` both do:

```python
from graphrag_sdk.core.providers.base import (
    _validate_timeout,
    _wait_for_provider_call,
)
```

Reads as a smell every time. Either move to a `providers/_timeout.py` private module (cross-module use stays internal to the package), or drop the underscore and treat them as a documented internal contract.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/multi_path.py
```diff
@@ -242,21 +251,29 @@ async def _execute(
...
+        ctx.ensure_budget("MultiPath relationship expansion")
         entity_list = list(found_entities.items())[: self._max_entities]
         relationship_strings = await expand_relationships(
-            self._graph, entity_list, self._max_relationships
+            self._graph, entity_list, self._max_relationships, ctx=ctx
         )
         ctx.log(f"MultiPath [5/9]: {len(relationship_strings)} relationships")
 
         # 6. Chunk retrieval (4 paths)
+        ctx.ensure_budget("MultiPath chunk retrieval")
```
galshubeli: 🟠 **H2 — Duplicate phase guard with inner check.**

This fires `ctx.ensure_budget('MultiPath chunk retrieval')` immediately before `retrieve_chunks(...)`, which itself runs `ctx.ensure_budget('chunk fulltext search')` microseconds later. The outer check changes the failure *label*, nothing else. Same pattern at lines 200, 238, 268, 293, 301, 323.

Pick one layer: trust the inner checks (cheaper to maintain, strategies are reusable) and drop the orchestrator duplicates, or keep orchestrator-only and strip the helpers. ~30 lines of churn collapse to zero either way.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/chunk_retrieval.py
```diff
@@ -17,6 +20,7 @@ async def retrieve_chunks(
     llm_kw: list[str],
     simple_kw: list[str],
     entity_list: list[tuple[str, dict]],
+    ctx: Context | None = None,
```
galshubeli, line 23: 🟠 **H3 — `ctx: Context | None = None` makes budget gating silently opt-in.**

Every retrieval helper takes `ctx` as optional and gates with `if ctx is not None: ctx.ensure_budget(...)`. A future caller that forgets to pass `ctx` silently disables the feature for that path with no warning.

Two cleaner options:
```python
# (a) make it required
async def retrieve_chunks(..., ctx: Context) -> ...:

# (b) use a module-level no-budget sentinel
_UNBOUNDED = Context()  # latency_budget_ms=None → ensure_budget is no-op
async def retrieve_chunks(..., ctx: Context = _UNBOUNDED) -> ...:
```

(a) is the mechanical-but-correct option.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/multi_path.py
```diff
@@ -186,33 +187,40 @@ async def _execute(
...
         if self._enable_cypher:
             results = await asyncio.gather(
-                search_relates_edges(self._vector, query_vector, self._rel_top_k),
-                execute_cypher_retrieval(self._graph, self._llm, query),
+                search_relates_edges(self._vector, query_vector, self._rel_top_k, ctx=ctx),
+                execute_cypher_retrieval(self._graph, self._llm, query, ctx=ctx),
                 return_exceptions=True,
             )
             # Unpack RELATES results
+            if isinstance(results[0], LatencyBudgetExceededError):
```
galshubeli: 🟠 **H4 — Fragile gather() unpacking.**

```python
if isinstance(results[0], LatencyBudgetExceededError): raise results[0]
if isinstance(results[0], BaseException): fact_strings_scored, rel_entities = [], {}
...
if isinstance(results[1], LatencyBudgetExceededError): raise results[1]
if not isinstance(results[1], BaseException): ...
```

If anyone adds a third coroutine, they have to remember the budget check or silently swallow it. Extract once:

```python
def _unpack(result, default):
    if isinstance(result, LatencyBudgetExceededError):
        raise result
    if isinstance(result, BaseException):
        return default
    return result
```

graphrag_sdk/src/graphrag_sdk/core/providers/litellm.py
```diff
@@ -183,11 +210,13 @@ async def ainvoke_messages(
         messages: list[ChatMessage],
         *,
         max_retries: int = 3,
+        timeout: float | None = None,
         **kwargs: Any,
     ) -> LLMResponse:
         """Native multi-turn completion via LiteLLM."""
         if max_retries < 1:
             raise ValueError("max_retries must be >= 1")
+        _validate_timeout(timeout)
```
galshubeli: 🟡 **M1 — Timeout validation drift between base and providers.**

`LiteLLM.ainvoke_messages` and `OpenRouterLLM.ainvoke_messages` call `_validate_timeout(timeout)` directly. But `base.LLMInterface.ainvoke_messages` doesn't — it delegates to `ainvoke` and lets validation happen there.

```python
# base.py — no direct validate
return await self.ainvoke(prompt, max_retries=max_retries, timeout=timeout, **kwargs)

# litellm.py — validates directly
_validate_timeout(timeout)
```

Either lift to base (and drop here), or push down consistently. Inconsistency is a footgun for the next override.

graphrag_sdk/src/graphrag_sdk/core/providers/base.py
```diff
@@ -160,7 +229,12 @@ async def ainvoke_messages(
...
         prompt = "\n\n".join(parts)
-        return await self.ainvoke(prompt, max_retries=max_retries, **kwargs)
+        return await self.ainvoke(
+            prompt,
+            max_retries=max_retries,
+            timeout=timeout,
+            **kwargs,
+        )
 
     async def astream(self, prompt: str, **kwargs: Any) -> AsyncIterator[str]:
```
galshubeli: 🟡 **M2 — `astream` has no `timeout=` parameter.**

```python
async def astream(self, prompt: str, **kwargs: Any) -> AsyncIterator[str]:
```

PR claims 'enforce LLM and embedding timeouts' but streaming is exempt. Streaming hangs are a real failure mode — a server can dribble tokens forever or stall mid-stream. Either expose `timeout=` here (and apply per-chunk), or document the exemption.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -1675,7 +1692,7 @@ async def _write_graph_config(self) -> None:
         except Exception:
             logger.debug("Failed to write graph config node", exc_info=True)
 
-    async def _validate_graph_config(self) -> None:
+    async def _validate_graph_config(self, ctx: Context | None = None) -> None:
```
galshubeli: 🟡 **M3 — Positional `ctx` adds subclass-break risk.**

```python
async def _validate_graph_config(self, ctx: Context | None = None) -> None:
```

This is a private method by name but it's on a public class. Any subclass overriding `_validate_graph_config(self)` will break silently if a future caller passes `ctx` positionally. Cheap fix:

```python
async def _validate_graph_config(self, *, ctx: Context | None = None) -> None:
```

Keyword-only matches the rest of the codebase's `*, ctx=...` convention.

graphrag_sdk/tests/test_facade.py
```diff
@@ -410,6 +449,87 @@ async def rerank(self, query, result, ctx):
...
+        mock_strategy = MagicMock(spec=RetrievalStrategy)
+        mock_strategy.search = AsyncMock(return_value=RetrieverResult())
+        g._retrieval_strategy = mock_strategy
+        ctx = Context(latency_budget_ms=1000.0)
+
+        def exhaust_budget(operation: str) -> None:
+            if operation == "graph config query":
+                raise LatencyBudgetExceededError("budget exhausted before config query")
+
+        ctx.ensure_budget = exhaust_budget  # type: ignore[method-assign]
```
galshubeli: 🟡 **M4 — Monkey-patching `ctx.ensure_budget` per-instance is brittle.**

```python
ctx.ensure_budget = exhaust_budget  # type: ignore[method-assign]
```

Also at `test_retrieval.py:179, 195` and similar pattern in `test_multi_path_retrieval.py`. If `ensure_budget` becomes a `cached_property` or moves to a parent class, these tests silently no-op (the patch lands on the instance, the call site reaches the class method).

Cleaner:
```python
monkeypatch.setattr(Context, 'ensure_budget', exhaust_budget)
# or a Context subclass with overridden behavior
```

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
```
galshubeli, line 11: 🟡 **M6 — Verify dependabot pip ecosystem works with PEP-621 `pyproject.toml`.**

```yaml
- package-ecosystem: "pip"
  directory: "/graphrag_sdk"
```

Dependabot's `pip` parser handles `[project.dependencies]` but has known gaps with `[project.optional-dependencies]` groups (it tends to only update `[project.dependencies]` and miss the `dev`/`extras` extras). Worth a dry run before relying on it, or switch to `uv` ecosystem if your `uv.lock` is the source of truth.

docker-compose.yml
```diff
@@ -3,13 +3,15 @@ services:
     image: falkordb/falkordb:v4.18.0
     ports:
       - "6379:6379"
+      - "3000:3000"
```
galshubeli: 🟡 **M7 — Browser UI binds to all interfaces by default.**

```yaml
ports:
  - "3000:3000"
```

Fine on a laptop, surprising on a shared dev host or CI runner — exposes the FalkorDB Browser UI on `0.0.0.0:3000`. Safer default:

```yaml
ports:
  - "127.0.0.1:3000:3000"
```

Or at minimum a one-line note in CONTRIBUTING.md so contributors aren't surprised.

.github/dependabot.yml
Naseem77: We didn’t fix M6 because this repo has no uv.lock, requirements.txt, or alternate dependency source to switch Dependabot to.

galshubeli (commented):
## Overview

The PR bundles six themes — broadly cohesive under "reliability." Most of the surface area is consistent and well-tested:

- New `LatencyBudgetExceededError`, `LLMTimeoutError`, `EmbeddingTimeoutError` types with proper hierarchy.
- `Context.ensure_budget(operation)` + `remaining_budget_seconds` plumbed through retrieval, completion, validation, and provider calls.
- Centralized `wait_for_provider_call` / `validate_timeout` helpers in `core/providers/_timeout.py`.
- Error logging upgraded from silent broad-except to `logger.error` + `logger.debug(exc_info=...)` in connection retry, retry helpers, loaders, pipeline, retrieval base.
- `-m integration` pytest marker isolates real-FalkorDB tests; CI runs them; CONTRIBUTING documents the local invocation.

## Issues

### 1. Duplicate comment lines in `multi_path.py` (merge artifact)
`graphrag_sdk/src/graphrag_sdk/retrieval/strategies/multi_path.py:234-293` has each step comment doubled:

```python
# 4. Entity discovery (2 paths) + merge rel_entities + cypher_entities
# 4. Entity discovery (2 paths) + merge rel_entities + cypher_entities
```

Same for steps 5, 6, 7, 8. Looks like rebase debris — should be cleaned up before merge.

### 2. Shared mutable `_UNBOUNDED = Context()` module-level singleton
Defined in `chunk_retrieval.py:13`, `entity_discovery.py:15`, `cypher_generation.py:19`, `relationship_expansion.py:13`, `result_assembly.py:17`. Today this works (because `latency_budget_ms=None` makes `ensure_budget` a no-op), but it's a footgun:

- `started_at` is set at import time; `elapsed_ms` grows forever.
- Any caller mutating `_UNBOUNDED.trace_id`, `tenant_id`, `latency_budget_ms` poisons every other caller.
- It's the classic "mutable default argument" anti-pattern, just hoisted to module scope.

Cleaner: `ctx: Context | None = None`, then `if ctx is not None: ctx.ensure_budget(...)`. Marginal verbosity is worth the safety.

### 3. Inconsistent timeout=0 behavior
Production path: `validate_timeout(0)` → `ValueError("timeout must be > 0")`.
Direct internal path (tested in `test_provider_wait_timeout_zero_raises_typed_error`): `wait_for_provider_call(..., timeout=0)` → typed `LLMTimeoutError`/`EmbeddingTimeoutError`.

The `if timeout <= 0` branch in `_timeout.py:18-22` is unreachable from production callers because every entry point calls `validate_timeout` first. The test exercises dead-but-buggy code. Pick one contract:

- If 0 is invalid → drop the branch in `wait_for_provider_call` and drop the test.
- If 0 means "fail-fast typed timeout" → remove `validate_timeout`'s zero check and let the helper own the semantics.

### 4. `remaining_budget_seconds` floor of `1e-9`
`context.py:55` clamps the returned timeout to `1e-9` seconds. Combined with `ensure_budget` immediately before each call, this is mostly safe — but for sub-millisecond remaining budgets you'll pass a clearly-infeasible timeout into `aembed_query`/`ainvoke` instead of failing fast. Consider raising `LatencyBudgetExceededError` if remaining is below a small threshold (e.g. 1 ms) rather than handing the LLM a 1-nanosecond deadline.

### 5. `wait_for_provider_call` closes arbitrary awaitables
`_timeout.py:18-22` does `getattr(awaitable, "close", None)` and calls it. Coroutines support `close()`, but generic `Awaitable` (futures, tasks) may not — and calling `close()` on a coroutine that hasn't been awaited prints a `RuntimeWarning` in some Python configurations. As noted above this branch is unreachable from production today, but if you keep it, narrow the type or use `inspect.iscoroutine`.

### 6. `LiteLLM` import: stray `import time` at top vs. lazy `import litellm`
`providers/litellm.py:9` adds `import time` at module level — fine — but `aembed_documents` (deadline path) is the only consumer. Just an FYI for consistency with the existing lazy-import pattern; not a bug.

## Smaller notes

- `pypi-publish.yaml`: `defaults.run.working-directory: graphrag_sdk` applies only to `run:` steps; `upload-artifact` and `pypa/gh-action-pypi-publish` see paths from repo root. `graphrag_sdk/dist/` is correct from repo root, so this works — but it's brittle. A comment would help.
- `_validate_history` swap from `try/except` on `ChatMessage(...)` to explicit type checks (`api/main.py:1488-1503`) is a good change — old code swallowed real bugs.
- `core/connection.py:194-208`: the \`if last_exc is not None\` block is written twice; the second one is needed but the first ends in \`raise last_exc\`. Reads slightly oddly — fine functionally.
- Async context manager tests reference `__aenter__`/`__aexit__` not in the diff. They exist in `api/main.py:213-237` already — verified, no missing implementation.
- Integration test discipline (`-m integration` marker + `RUN_INTEGRATION=1` doc) is well executed.

## Test coverage

Strong. New tests cover timeout propagation per provider, budget exhaustion at each retrieval phase, log-redaction sanity (`SECRET_KEY_xyz`/`proxy=` exclusion from error logs), context-manager close-failure behavior, and the `aembed_documents` overall-deadline guarantee.

## Risk assessment

**Medium-low.** The semantic change with most blast radius is making every retrieval step budget-aware — well-behaved when no budget is set (no-op), but callers passing tight budgets will now see `LatencyBudgetExceededError` mid-flow rather than silent timeouts. This is intentional and an improvement, but is API-visible.

## Recommendation

Request changes for items 1 (cleanup) and ideally 2 (sentinel pattern). Items 3–5 are worth addressing but non-blocking. The bulk of the work is solid.

galshubeli (commented):
## Follow-up review — commit 4fc166b

Verified each item from the previous review against the follow-up commit.

### Resolved

| # | Issue | Status |
|---|-------|--------|
| 1 | Duplicate comment lines in `multi_path.py` | ✅ Removed at steps 4–8 |
| 2 | Shared mutable `_UNBOUNDED = Context()` singleton across 5 retrieval modules | ✅ Replaced with `ctx: Context \| None = None` + per-call-site `if ctx is not None: ctx.ensure_budget(...)` guards |
| 3 | Inconsistent `timeout=0` behavior between `validate_timeout` (ValueError) and `wait_for_provider_call` (typed timeout) | ✅ Dead `if timeout <= 0` branch removed from `_timeout.py`; `validate_timeout` is now the single source of truth; dead-code test removed |
| 4 | `remaining_budget_seconds` 1e-9 clamp would hand the provider a sub-millisecond deadline | ✅ Clamp removed; new `Context.provider_timeout_seconds(operation, min_remaining_ms=1.0)` raises `LatencyBudgetExceededError` if remaining ≤ 1 ms, otherwise returns the seconds value. All budget-derived provider timeouts now route through it. |
| 5 | `wait_for_provider_call` calling `close()` on arbitrary awaitables | ✅ Eliminated with the `timeout <= 0` branch |
| — | Doubled `if last_exc is not None:` block in `connection.py` | ✅ Consolidated into a single guard wrapping both `logger.debug` and `raise last_exc` |

New tests in `test_context.py` cover both the raise-when-low and return-remaining paths of `provider_timeout_seconds`. Good coverage.

### Minor follow-up (non-blocking)

`provider_timeout_seconds` itself raises `LatencyBudgetExceededError` when remaining ≤ 1 ms, so the immediately-preceding `ctx.ensure_budget(operation)` call is redundant at every site that pairs them — e.g. `local.py:55-58`, `cosine.py:43-46`, `api/main.py:1527-1532` and `api/main.py:1656-1660`. The two have slightly different thresholds (`ensure_budget` fires at remaining ≤ 0, `provider_timeout_seconds` at remaining ≤ 1 ms), but since the latter is stricter, the prior call adds nothing — it can never fire when the next call wouldn't.

Two ways to clean up:
- Drop the redundant `ensure_budget(...)` and rely on `provider_timeout_seconds(...)` raising.
- Or keep `ensure_budget` only for the call sites that don't pass a timeout (graph queries, vector searches), and drop it elsewhere.

Either way, current behavior is correct — just a small simplification opportunity.

### Recommendation

LGTM to merge. Optional polish above can be done in a follow-up or skipped.

galshubeli (approved):

PR #262, opened 2026-05-19
feat(loaders): unify ingestion via Docling (DOCX/XLSX/PPTX/HTML/CSV + URL)

### LINKED ISSUE #241: Expand document loader coverage (HTML, Markdown, DOCX, CSV, JSON, URL, S3, Image)
## Problem

Only Text and PDF loaders ship today. Anyone with HTML, Markdown, DOCX, CSV/Excel, JSON, URL-based, S3-hosted, or image content has to write a custom `LoaderStrategy` before they can try the SDK.

## Goal

Broaden built-in datatype support so common formats work out of the box.

## Roadmap

Phase 1.1 (Document Loaders) and Phase 1.3 (Structured Data to Graph). Marked :handshake: — good for community contributions.

drr00t:
## Summary
This feature extends the loader interface using Docling to add support for DOCX, XLSX, PPTX, CSV, HTML, and XHTML. 

## Changes
- **DoclingBaseLoader**: Base class wrapping the Docling document model.

## Test Plan
- [X] All existing tests pass (`pytest tests/ -q`)
- [X] New tests added for new functionality (if applicable)
- [X] Lint passes (`ruff check src/`)

## Notes
Currently, the `GraphRAG_SDK` loader interface maps one loader to one extension. Since Docling supports multiple extensions, this PR updates the current loader implementation paradigm to support multi-format loaders. 

## Summary by CodeRabbit

* **New Features**
  * Add optional Docling-based universal loader for DOCX/XLSX/PPTX/HTML/CSV and URLs (installable via an optional "docling" extra). Automatic fallback to this loader for those formats when installed.

* **Documentation**
  * Updated docs and README with installation note, supported formats, strategy reference, and usage examples.

* **Examples**
  * Added an example showing advanced Docling loader usage.

* **Tests**
  * Added comprehensive tests covering Docling ingestion behaviors and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FalkorDB/GraphRAG-SDK/pull/262?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


galshubeli (commented):
_(Withdrawn — will repost after internal review.)_

galshubeli (commented):
Hi @drr00t — first, thank you for tackling this. Issue #241 has been open for a while and you've done the hard part: introducing docling, wiring up the dispatcher, adding tests, and getting the integration working end-to-end. The direction is right and the foundation is solid.

After looking at this more carefully, the main thing I'd like to discuss is the architecture. I think there's a meaningfully simpler shape here.

## 🎯 Headline change: one universal loader, not five

Docling's `DocumentConverter` already auto-detects format from the source — one converter object handles PDF, DOCX, XLSX, PPTX, HTML, CSV, Markdown, URLs, and more. The PR currently wraps that single object in **five empty subclasses** that differ only by an `extension_name` string, and adds **five new branches** to `_default_loader_for`. Leaning into docling's unified API collapses both.

**Before** (this PR):
```python
# 5 loader files, ~60 LOC of boilerplate
class CsvLoader(DoclingBaseLoader):  extension_name = "csv"
class DocxLoader(DoclingBaseLoader): extension_name = "docx"
class HtmlLoader(DoclingBaseLoader): extension_name = "html"
class PptxLoader(DoclingBaseLoader): extension_name = "pptx"
class XlsxLoader(DoclingBaseLoader): extension_name = "xlsx"

# _default_loader_for, 5 new branches
if lower.endswith(".docx"): return DocxLoader()
if lower.endswith(".xlsx"): return XlsxLoader()
if lower.endswith(".pptx"): return PptxLoader()
if lower.endswith(".html") or lower.endswith(".xhtml"): return HtmlLoader()
if lower.endswith(".csv"):  return CsvLoader()
```

**After**:
```python
class DoclingLoader(LoaderStrategy):
    """Universal loader. Delegates to docling for any supported format."""
    def __init__(self, **docling_kwargs): ...
    async def load(self, source: str, ctx: Context) -> DocumentOutput:
        return await asyncio.to_thread(self._load_sync, source)

# Dispatcher collapses to a single line — docling handles everything.
def _default_loader_for(source: str) -> LoaderStrategy:
    return DoclingLoader()
```

Why this is better:
- **5 classes → 1**, **5 dispatch branches → 1 line**.
- **All formats through one path** — PDF, DOCX, XLSX, PPTX, HTML, CSV, MD, URL — no extension sniffing, no branching.
- **URLs work for free** — docling accepts URLs as `source`, partially closing the URL line item in #241.
- **Future formats are free** — when docling adds AsciiDoc, JATS, USPTO XML, etc., GraphRAG-SDK picks them up automatically.
- **Format-specific kwargs still possible** via docling's `format_options` dict if needed later.
- **Existing loaders stay available** — `TextLoader`, `PdfLoader`, `MarkdownLoader` aren't removed; users who want them can pass `loader=PdfLoader()` explicitly. The change is only in what the *default* is.

## Must-fix

Nine stray scratch files in the diff — looks like leftovers from working out the mock-import setup:

```
test_import_mock.py             (repo root)
test_mock_import.py             (repo root)
test_patch_dict.py              (repo root)
test_sys_modules.py             (repo root)
graphrag_sdk/test_debug.py
graphrag_sdk/test_monkeypatch.py
graphrag_sdk/test_monkeypatch2.py
graphrag_sdk/test_monkeypatch3.py
graphrag_sdk/dummy_path          (0-byte empty file)
```

The working version made it into `tests/test_docling_loaders.py`; these can all be deleted.

## Smaller polish

- PR title: consider `feat(loaders): unify ingestion via Docling (DOCX/XLSX/PPTX/HTML/CSV + URL)` (the current title is just the branch name).
- In `docling_base.py`, the `metadata["label"]` fallback `label.value if hasattr(label, "value") else label` can store a non-serializable enum object — coerce with `str(label)` so downstream storage doesn't break.
- With the unified `DoclingLoader`, one smoke test per format (CSV header row, HTML script/style stripping, XLSX multi-sheet, MD breadcrumbs) would catch most regressions.
- README/CONTRIBUTING update listing the now-supported formats would help discoverability.

## Scope

Issue #241 asks for HTML, Markdown, DOCX, CSV, JSON, URL, S3, Image. With the unified `DoclingLoader` this PR covers HTML, DOCX, CSV, XLSX, PPTX, Markdown, **and URL** in one shot — and PDF goes through the same path too. JSON, S3, and Image can land in follow-ups (S3 in particular is its own thing — a source adapter rather than a format).

Happy to chat in the thread if anything here doesn't land. Thanks again for the contribution!

drr00t:
Hi @galshubeli, thank you very much for the review. I agree with all your comments. When looking into Docling, I noticed that graphrag_sdk didn't have the concept of multi-loader documents. Although we hadn't discussed it, I decided to implement it to see if it makes sense. I tried to minimize the impact by creating a "fake" DoclingBaseLoader to avoid duplicating too much code. 

I will make the requested changes.

drr00t:
@galshubeli, it's good to go.

galshubeli (commented):
Walked the loader, tests, and wiring. Listing findings in priority order.

**1. URL support is broken (`docling_loader.py:43`)**
The docstring says DoclingLoader handles URLs, but `Path(source).exists()` is always False for `https://...`, so URLs raise `LoaderError("File not found")` before docling ever sees them. Either gate the existence check on URL scheme, or remove the URL claim from the docstring and `docs/strategies.md:69`.

**2. Default loader silently changed (`api/main.py` `_default_loader_for`)**
The unknown-extension fallback was `TextLoader()`; this PR makes it `DoclingLoader()`. Someone on `pip install graphrag-sdk[litellm]` (no docling) who calls `rag.ingest("notes.rst")` used to get it read as text — now gets a runtime "this format requires docling" error. Suggest dispatching DoclingLoader only for the explicit format list (.docx/.xlsx/.pptx/.html/.xhtml/.csv) and keeping TextLoader as the catch-all.

**3. Tables probably come out empty (`docling_loader.py:110-117`)**
Docling's `TableItem` typically doesn't set `.text` — the table content lives in `.data` or comes out via `export_to_markdown()`. The code does `getattr(item, "text", "")` first. The test passes only because it mocks a fake TableItem with `.text="Table content"` — not how real docling works. Worth verifying with a real .docx that has a table.

**4. Bare `except: pass` (`docling_loader.py:71-72`)**
If `export_to_markdown()` fails for any real reason, the item disappears silently with no log. Just needs a debug log line.

**5. `docling` in the `all` extra (`pyproject.toml:67`)**
Docling pulls torch (~800 MB) + EasyOCR weights, so `pip install graphrag-sdk[all]` goes from a few hundred MB to gigabytes. `pdf-fast` (PyMuPDF) is deliberately kept out of `all` for the AGPL reason — same precedent could apply here on footprint grounds.

**6. Tests don't actually run docling (`tests/test_docling_loaders.py:34-57`)**
The autouse fixture replaces `docling.document_converter` with a MagicMock for every test. If docling renames `iterate_items()` or drops `DocItemLabel.TABLE`, the suite still passes but users break. At least one `pytest.importorskip("docling")` test against a tiny real fixture would catch this class of regression.

---

Severity:
- #1, #2 — block merge (broken feature + behavior regression)
- #3 — likely broken in practice, needs verification on real input
- #4, #5, #6 — should address but don't block

Happy to discuss any of these.

drr00t:
Hi @galshubeli, All items were addressed, I had to reformat multiple files.

PR #274, opened 2026-06-22
feat: agentic retrieval - graph-walk, skills, MCP server, and MultiPath path router

Naseem77:
## Summary

Phase-3 work adding **agentic retrieval** on top of the existing MultiPath pipeline, plus supporting graph algorithms, a skills framework, an MCP server, and a new **agentic path router**.

Split into 10 focused commits (dependency-ordered; every commit is importable, exports are wired last).

## What's included

- **core models** — `AgentStep`, `AgentTrace`, `ScoredPath`, `SkillResult`.
- **storage** — `GraphStore.pagerank()` and `weighted_neighbors()`.
- **graph-walk** — `DynamicGraphWalk` (beam + bidirectional) and `GraphWalkRetrieval`.
- **agentic loop** — `AgenticRetrieval`, a ReAct loop where the LLM picks tools (`search`, `cypher`, `traverse`, + skills) via a guard-railed `ToolRegistry`.
- **skills** — `Skill` base + registry and 5 built-ins (entity_comparison, impact_analysis, contradiction_detection, gap_analysis, timeline_reconstruction).
- **api** — `GraphRAG.run_skill(name, **params)`.
- **mcp** — MCP server exposing retrieval + skills, `graphrag-mcp` console script, optional `[mcp]` extra.
- **path router** — opt-in agentic selection of which of MultiPath's 5 paths (`relates`, `entity_cypher`, `entity_fulltext`, `expansion`, `chunks`) to run per question. `LLMPathRouter` + `HeuristicPathRouter`, with all-paths fallback on empty/error/timeout so recall is never lost. Default behavior unchanged.
- **exports** — Phase-3 public API re-exported from `graphrag_sdk`.

## Testing

New unit tests: `test_graph_walk.py`, `test_agentic_retrieval.py`, `test_skills.py`, `test_mcp_tools.py`, `test_path_router.py`.

Full suite: **1083 passed, 29 skipped** (1 pre-existing, unrelated LiteLLM timing test failure; no provider code touched here).

## Summary by CodeRabbit

* **New Features**
  * Added Model Context Protocol (MCP) support: `graphrag-mcp` CLI plus an MCP server with stdio and SSE transports.
  * Introduced agentic (ReAct-style) retrieval with a tool registry and stricter read-only Cypher handling.
  * Added graph-walk retrieval with weighted path scoring (beam and bidirectional search).
  * Implemented retrieval path routing and plan-driven MultiPathRetrieval stage gating.
  * Added skills framework, built-in skills, top-level `run_skill()` support, and expanded SDK exports.
* **Bug Fixes**
  * Added configurable non-negative result caps via updated result assembly (`max_*` limits).
* **Tests**
  * Added new test coverage for agentic retrieval, graph walk, MCP tools, path routing, and built-in skills.

### DIFF
--- a/graphrag_sdk/src/graphrag_sdk/mcp/tools.py
+++ b/graphrag_sdk/src/graphrag_sdk/mcp/tools.py
@@ -1,0 +1,218 @@
+# GraphRAG SDK — MCP: Tool definitions (Phase 3.2)
+# Transport-agnostic tool specs that wrap a GraphRAG facade. Kept free of
+# any `mcp` package import so they can be unit-tested without the optional
+# dependency installed; server.py adapts these into a live MCP server.
+
+from __future__ import annotations
+
+import json
+import logging
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass, field
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+MCPHandler = Callable[[dict[str, Any]], Awaitable[str]]
+
+
+@dataclass
+class MCPTool:
+    """A single MCP tool: name, description, JSON input schema, and handler."""
+
+    name: str
+    description: str
+    input_schema: dict[str, Any]
+    handler: MCPHandler
+
+    def to_spec(self) -> dict[str, Any]:
+        return {
+            "name": self.name,
+            "description": self.description,
+            "inputSchema": self.input_schema,
+        }
+
+
+def _obj(props: dict[str, Any], required: list[str] | None = None) -> dict[str, Any]:
+    return {
+        "type": "object",
+        "properties": props,
+        "required": required or [],
+    }
+
+
+@dataclass
+class GraphRAGToolset:
+    """Builds the standard 8-tool MCP surface over a GraphRAG instance."""
+
+    rag: Any
+    tools: list[MCPTool] = field(default_factory=list)
+
+    def __post_init__(self) -> None:
+        self.tools = self._build()
+
+    def by_name(self, name: str) -> MCPTool | None:
+        for tool in self.tools:
+            if tool.name == name:
+                return tool
+        return None
+
+    def specs(self) -> list[dict[str, Any]]:
+        return [t.to_spec() for t in self.tools]
+
+    def _build(self) -> list[MCPTool]:
+        rag = self.rag
+
+        async def ingest(args: dict[str, Any]) -> str:
+            result = await rag.ingest(
+                args["text"], document_id=args.get("document_id")
+            )
+            return _dump(
+                {
+                    "nodes_created": getattr(result, "nodes_created", None),
+                    "relationships_created": getattr(result, "relationships_created", None),
+                    "chunks_indexed": getattr(result, "chunks_indexed", None),
+                }
+            )
+
+        async def retrieve(args: dict[str, Any]) -> str:
+            result = await rag.retrieve(args["question"])
+            return _dump({"items": [i.content for i in result.items]})
+
+        async def answer(args: dict[str, Any]) -> str:
+            result = await rag.completion(args["question"])
+            return _dump({"answer": result.answer, "metadata": result.metadata})
+
+        async def cypher_query(args: dict[str, Any]) -> str:
+            from graphrag_sdk.retrieval.agentic.tools import is_read_only_cypher
+
+            cypher = args["cypher"]
+            if not is_read_only_cypher(cypher):
+                return "Error: only read-only Cypher is permitted via MCP."
+            result = await rag._graph_store.query_raw(cypher)
+            rows = list(getattr(result, "result_set", []) or [])
+            return _dump({"rows": [[_jsonable(c) for c in row] for row in rows]})
+
+        async def graph_walk(args: dict[str, Any]) -> str:
+            from graphrag_sdk.retrieval.graph_walk import DynamicGraphWalk
+
+            store = rag._graph_store
+            try:
+                weights = await store.pagerank()
+            except Exception:
+                weights = {}
+
+            async def neighbor_fn(node_id: str) -> list[tuple[str, float, str]]:
+                return await store.weighted_neighbors(node_id)
+
+            walk = DynamicGraphWalk(
+                neighbor_fn,
+                node_weights=weights,
+                beam_width=int(args.get("beam_width", 5)),
+                max_depth=int(args.get("max_depth", 3)),
+            )
+            goal = args.get("goal")
+            if goal:
+                path = await walk.bidirectional_search(args["start"], str(goal))
+                return _dump({"path": path.model_dump() if path else None})
+            paths = await walk.beam_search(args["start"])
+            return _dump({"paths": [p.model_dump() for p in paths]})
+
+        async def run_skill(args: dict[str, Any]) -> str:
+            from graphrag_sdk.skills import build_skill
+
+            skill = build_skill(args["skill"], rag._graph_store, rag.llm)
+            params = args.get("params", {}) or {}
+            result = await skill.run(None, **params)
+            return _dump(result.model_dump())
+
+        async def get_statistics(_args: dict[str, Any]) -> str:
+            return _dump(await rag.get_statistics())
+
+        async def get_ontology(_args: dict[str, Any]) -> str:
+            ontology = await rag.get_ontology()
+            return _dump(ontology.model_dump() if hasattr(ontology, "model_dump") else {})
+
+        return [
+            MCPTool(
+                "ingest",
+                "Ingest raw text into the knowledge graph.",
+                _obj(
+                    {
+                        "text": {"type": "string"},
+                        "document_id": {"type": "string"},
+                    },
+                    ["text"],
+                ),
+                ingest,
+            ),
+            MCPTool(
+                "retrieve",
+                "Retrieve graph context for a question (no generation).",
+                _obj({"question": {"type": "string"}}, ["question"]),
+                retrieve,
+            ),
+            MCPTool(
+                "answer",
+                "Full RAG pipeline: retrieve context and generate an answer.",
+                _obj({"question": {"type": "string"}}, ["question"]),
+                answer,
+            ),
+            MCPTool(
+                "cypher_query",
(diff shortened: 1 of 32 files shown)


graphrag_sdk/src/graphrag_sdk/mcp/tools.py
```diff
@@ -0,0 +1,216 @@
...
+        return None
+
+    def specs(self) -> list[dict[str, Any]]:
+        return [t.to_spec() for t in self.tools]
+
+    def _build(self) -> list[MCPTool]:
+        rag = self.rag
+
+        async def ingest(args: dict[str, Any]) -> str:
+            result = await rag.ingest(args["text"], document_id=args.get("document_id"))
```
galshubeli: **MCP `ingest` can never ingest text.** `GraphRAG.ingest(source, *, text=...)` treats the first positional arg as a **file path** — raw text must go through `text=`. As written, an MCP client calling `ingest` with `{"text": "..."}` makes the loader try to open the prose as a path and the tool returns `Error running tool 'ingest'`.

```suggestion
            result = await rag.ingest(text=args["text"], document_id=args.get("document_id"))
```

graphrag_sdk/src/graphrag_sdk/storage/graph_store.py
```diff
@@ -331,6 +331,72 @@ async def query_raw(self, cypher: str, params: dict[str, Any] | None = None) ->
...
+    ) -> list[tuple[str, float, str]]:
+        """Return weighted neighbors of an entity for dynamic graph walks.
+
+        Each tuple is ``(neighbor_id, edge_weight, edge_label)``. Edge
+        weight falls back to ``1.0`` when the relationship has no stored
+        ``weight`` property.
+        """
+        query = (
+            "MATCH (n:__Entity__ {id: $node_id})-[r]-(m:__Entity__) "
+            "RETURN m.id AS id, type(r) AS label, "
```
galshubeli: **`type(r)` collapses every edge to `"RELATES"`.** The schema stores a single `RELATES` edge type with the actual relation in the `r.rel_type` property (see `relationship_expansion.py:23,42`). Using `type(r)` here means every `ScoredPath.edges` label and `traverse` result reads `"RELATES"` instead of `BORN_IN`/`WORKS_AT`/etc., silently losing the semantic relation.

```suggestion
            "RETURN m.id AS id, coalesce(r.rel_type, type(r)) AS label, "
```

graphrag_sdk/src/graphrag_sdk/skills/contradiction_detection.py
```diff
@@ -0,0 +1,79 @@
...
+        entity = params.get("entity")
+        if not entity:
+            raise ValueError("contradiction_detection requires 'entity'")
+        limit = int(params.get("limit", 50))
+        if limit < 1:
+            raise ValueError("contradiction_detection 'limit' must be >= 1")
+
+        rows = await self._rows(
+            "MATCH (e:__Entity__ {id: $id})-[r]-(m:__Entity__) "
+            "RETURN type(r) AS rel, m.id AS other, "
```
galshubeli: Same `type(r)` issue as in `graph_store.weighted_neighbors`: every fact line becomes `"RELATES <other>"` instead of the semantic relation, so a contradiction that hinges on the relation type (e.g. `BORN_IN Paris` vs `BORN_IN London`) is invisible to the LLM. Prefer `coalesce(r.rel_type, type(r)) AS rel`.

graphrag_sdk/src/graphrag_sdk/retrieval/agentic/loop.py
```diff
@@ -0,0 +1,180 @@
...
+    RetrieverResultItem,
+)
+from graphrag_sdk.retrieval.agentic.prompts import render_system_prompt
+from graphrag_sdk.retrieval.agentic.tools import ToolRegistry, build_default_registry
+from graphrag_sdk.retrieval.strategies.base import RetrievalStrategy
+
+logger = logging.getLogger(__name__)
+
+_ACTION_RE = re.compile(r"Action\s*:\s*(.+)", re.IGNORECASE)
+_ACTION_INPUT_RE = re.compile(r"Action\s*Input\s*:\s*(\{.*\})", re.IGNORECASE | re.DOTALL)
```
galshubeli: **Greedy `Action Input` capture drops valid tool args.** `(\{.*\})` with `DOTALL` spans from the first `{` to the *last* `}` in the whole response. When the model emits anything containing `}` after the JSON (a trailing `Observation:` line, prose), `json.loads` fails → `action_input = {}` → the `search` tool then returns `"Error: 'query' is required."` Verified locally with a trailing `Observation: {...}`. Consider a non-greedy / brace-balanced parse.

graphrag_sdk/src/graphrag_sdk/retrieval/agentic/tools.py
```diff
@@ -0,0 +1,273 @@
...
+
+logger = logging.getLogger(__name__)
+
+ToolHandler = Callable[[dict[str, Any], Context], Awaitable[str]]
+
+# Cypher write/DDL keywords rejected by the read-only cypher tool. Matched on
+# word boundaries so substrings of legitimate identifiers (e.g. "recall",
+# "asset") are not falsely flagged. ``call`` blocks every stored-procedure path
+# (algo.*, db.*, dbms.*, apoc.*) since procedures can mutate the graph.
+_WRITE_KEYWORD_RE = re.compile(
```
galshubeli, line 25: **Read-only gate false-positively rejects valid queries.** `_WRITE_KEYWORD_RE` matches write keywords anywhere in the raw text, including inside string literals. Verified: `MATCH (n) WHERE n.name = 'call center' RETURN n` and `... CONTAINS "delete" ...` are both rejected as non-read-only (by the agent *and* the MCP `cypher_query` tool). Note the codebase already has `validate_cypher()` in `cypher_generation.py` — worth reusing rather than maintaining a second allow/deny list.

graphrag_sdk/src/graphrag_sdk/skills/impact_analysis.py
```diff
@@ -0,0 +1,78 @@
...
+            for hop, node in enumerate(path.nodes[1:], start=1):
+                existing = impacted.get(node)
+                if existing is None:
+                    impacted[node] = {
+                        "distance": hop,
+                        "score": path.score,
+                        "via": path.nodes,
+                    }
+                    continue
+                if hop < existing["distance"]:
```
galshubeli: **`score` and `distance`/`via` can come from different paths.** `distance`/`via` update on `hop < existing["distance"]` while `score` updates independently on `path.score > existing["score"]`. A node first seen at depth 2 (score 0.9) then depth 1 (score 0.3) ends up `distance=1, score=0.9` — a mismatched pair that the `(distance, -score)` ranking key then sorts on. Update `score`/`via` together with `distance`.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/multi_path.py
```diff
@@ -216,6 +222,28 @@ async def _execute(
...
     ) -> RawSearchResult:
+        # 0. Resolve per-call overrides. Defaults preserve normal-flow
+        # behavior; the agentic flow may pass kwargs to widen selection.
+        def _pos_int(key: str, default: int) -> int:
+            val = kwargs.get(key, default)
+            try:
+                val = int(val)
+            except (TypeError, ValueError):
+                return default
+            return val if val > 0 else default
```
galshubeli, line 233: **`_pos_int` coerces `0` to the default, so the documented "0 = omit section" is unreachable.** `result_assembly.assemble_raw_result` clamps negatives and documents `0 = omit it entirely`, but every value reaches it through this `return val if val > 0 else default`. An agent passing `max_passages_out=0` to drop a section silently gets the full 15; same for the breadth knobs (`chunk_top_k=0`, etc.). Allow `>= 0` for the output caps.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/path_router.py
```diff
@@ -0,0 +1,114 @@
...
+        plan: set[str] = {"relates", "chunks"}
+        # Relationship / connection questions → traversal.
+        rel_pattern = (
+            r"\b(connect|related|relationship|between|link|neighbor"
+            r"|path|how (?:are|is|do))\b"
+        )
+        if re.search(rel_pattern, q):
+            plan |= {"expansion", "entity_cypher"}
+        # Quoted or capitalized proper nouns in the original query → name match.
+        if '"' in query or "'" in query or re.search(r"\b[A-Z][a-z]+\b", query or ""):
```
galshubeli: **Heuristic rarely prunes the entity paths.** `re.search(r"\b[A-Z][a-z]+\b", query)` matches the sentence-cased leading word of nearly every question ("What…", "How…"), so `entity_cypher` + `entity_fulltext` get added on almost all queries — undercutting the router's stated purpose of skipping irrelevant paths. Consider excluding a leading stop-word / requiring a capitalized token that isn't the first word.

graphrag_sdk/src/graphrag_sdk/retrieval/strategies/multi_path.py
```diff
@@ -283,22 +348,28 @@ async def _execute(
                     f"({len(found_entities)} total)"
                 )
         # 5. Relationship expansion
-        entity_list = list(found_entities.items())[: self._max_entities]
-        relationship_strings = await expand_relationships(
-            self._graph, entity_list, self._max_relationships, ctx=ctx
-        )
+        entity_list = list(found_entities.items())[:max_entities]
+        if "expansion" in plan:
```
galshubeli, line 372: **A router plan of `expansion` alone yields zero relationships.** `expansion` is gated independently of the paths that populate `entity_list`. `parse_paths` accepts `{"expansion"}` with no entity path, leaving `entity_list` empty so `expand_relationships([], …)` returns `[]` — the exact data the path exists to provide is dropped on a "how are X and Y connected" question. `HeuristicPathRouter` couples them, but nothing in the strategy enforces it; consider auto-including an entity path whenever `expansion` is selected.

PR #275, opened 2026-07-02
feat(ingestion): agentic per-document strategy selection

Naseem77:
## What

Adds **agentic ingestion strategy selection** to `GraphRAG.ingest(...)`, mirroring the retrieval-side `PathRouter`. When `auto=True`, a planner inspects a sample of each document and picks the **chunker**, **entity-extraction backend**, and **resolver** per document — filling in only the strategies the caller did *not* pass explicitly.

## Design

- Explicit `chunker` / `extractor` / `resolver` always win.
- On an empty/invalid plan or any planner error, it falls back to today's defaults (`sentence` / `gliner` / `exact`), so behavior is never silently lost.
- Two planners:
  - `LLMIngestionPlanner` (default) — one small guided LLM call.
  - `HeuristicIngestionPlanner` — zero-cost, feature-based.
- The planner may also tune parameters inside each strategy; all values are coerced and clamped to safe ranges (`PARAM_SPECS`), so the model can never produce an unsafe config.

## Usage

```python
# LLM planner picks strategies per document
await rag.ingest(source="doc.md", auto=True)

# zero-cost heuristic instead
from graphrag_sdk import HeuristicIngestionPlanner
await rag.ingest(source="doc.md", auto=True, planner=HeuristicIngestionPlanner())

# explicit strategy still wins; planner fills the rest
await rag.ingest(source="doc.md", auto=True, resolver=ExactMatchResolution())
```

## Commits

1. `IngestionPlan` schema + param clamping
2. Strategy builders
3. Heuristic + LLM planners
4. Package exports
5. `auto=`/`planner=` wiring in `GraphRAG.ingest`
6. Tests: schema, parsing, tunable params
7. Tests: planners, builders, `ingest(auto=)` merge

## Tests

`38 passed` — `tests/test_ingestion_planner.py`. Ruff clean on all changed files.

## Summary by CodeRabbit

* **New Features**
  * Added automatic, per-document ingestion strategy selection during ingestion.
  * Exposed new ingestion-planning options at the top-level SDK for easier access.
  * Added support for choosing chunking, extraction, and resolution strategies based on document content or planner output.

* **Bug Fixes**
  * Improved fallback behavior so ingestion continues with default strategies when planning cannot be completed.
  * Added validation and safe parameter clamping for supported ingestion strategy settings.

### DIFF
--- a/docs/ingestion.md
+++ b/docs/ingestion.md
@@ -246,7 +246,7 @@ stats = await rag.finalize()
 
 ## Agentic Strategy Selection (`auto=True`)
 
-By default `ingest()` uses fixed strategies (see [Configuration Quick Reference](#configuration-quick-reference)). Pass `auto=True` to let a **planner** inspect a sample of each document and pick the chunker, entity-extraction backend, and resolver *per document* — the ingestion-side analogue of retrieval's path router.
+By default `ingest()` uses fixed strategies (see [Configuration Quick Reference](#configuration-quick-reference)). Pass `auto=True` to let a **planner** inspect a sample of each document and pick the chunker, entity-extraction backend, and resolver *per document* — the ingestion-side analogue of retrieval-side routing.
 
 ```python
 # LLM planner (default): one small call decides the strategies per document
--- a/docs/ingestion.md
+++ b/docs/ingestion.md
@@ -244,6 +244,49 @@ stats = await rag.finalize()
 
 ---
 
+## Agentic Strategy Selection (`auto=True`)
+
+By default `ingest()` uses fixed strategies (see [Configuration Quick Reference](#configuration-quick-reference)). Pass `auto=True` to let a **planner** inspect a sample of each document and pick the chunker, entity-extraction backend, and resolver *per document* — the ingestion-side analogue of retrieval's path router.
+
+```python
+# LLM planner (default): one small call decides the strategies per document
+await rag.ingest(source="handbook.md", auto=True)
+
+# Zero-cost heuristic planner instead of an LLM call
+from graphrag_sdk import HeuristicIngestionPlanner
+await rag.ingest(source="handbook.md", auto=True, planner=HeuristicIngestionPlanner())
+
+# Explicit strategies always win — the planner only fills the rest
+await rag.ingest(source="handbook.md", auto=True, resolver=ExactMatchResolution())
+```
+
+**What the planner may pick:**
+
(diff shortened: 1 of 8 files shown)


galshubeli:
## Review — agentic ingestion strategy selection

Nice work overall — the safety-first design is solid: strategy ids validated, params clamped with the cross-field guards (`chunk_overlap < chunk_size`, `hard > soft`) matching the real constructor constraints, and `LatencyBudgetExceededError` correctly re-raised while everything else degrades to defaults. Mirrors the retrieval `PathRouter` contract cleanly.

Issues found, by size:

**Big**
1. **Double document load** — `_plan_ingestion_strategies` fully loads the file to get a content sample, then `IngestionPipeline` loads it again. With `auto=True` that's two full parses per document (expensive for PDFs, multiplied in batch mode), and the sample load also runs for `HeuristicIngestionPlanner`, which mostly needs the path. Suggest caching the loaded document for the pipeline, or at least skipping the sample load when the planner doesn't need content and documenting the extra cost.

**Medium**

2. **No end-to-end test of `ingest(auto=True)`** — tests drive `_plan_ingestion_strategies` via a stub, but nothing exercises the real `ingest()` wiring (planner runs after loader resolution + ontology init). One integration test with a mocked LLM would close this.

**Small**

3. `planner: Any | None` in the public API — a small `Protocol` (`async def plan(...) -> IngestionPlan | None`) would document the contract and give type-checker support.
4. `parse_plan` partial-JSON gap — if the JSON contains any one of the three keys, the key:value fallback scrape is skipped for the others. Degrades safely; just noting.
5. Double clamping (in `IngestionPlan.__post_init__` and again in `build_*`) — harmless/idempotent, but a one-line comment would prevent a future "cleanup".
6. `_GUIDE` omits `semantic`'s `force_summary_threshold`/`max_summary_tokens` even though `PARAM_SPECS` allows them — either add to the guide or drop from the spec.
7. Prompt injection via document content can steer the planner; blast radius is well-bounded (valid strategies, clamped params — worst case a pricier config), worth one sentence in the docs' safety-guarantees list.
8. Docs link to `blob/main/ingestion_planner.py`, which only exists once this and #274 land on main — fine at merge time.

Also noting the PR is stacked on #274's branch, so merge order matters.

**Verdict: approve with minor changes** — #1 is the one I'd want addressed (or consciously accepted) before merge; the rest is polish.

PR #279, opened 2026-07-21
docs: migrate to v1.2 ontology vocabulary

galshubeli:
## Summary

The documentation, README, and runnable examples still used the pre-v1.2 class names — `GraphSchema` / `EntityType` / `RelationType` / `PropertyType` — and the deprecated `schema=` kwarg. Copy-pasting any docs example on a current SDK emits `DeprecationWarning`s, and the aliases are slated for removal.

This PR migrates everything to the current vocabulary (`Ontology` / `Entity` / `Relation` / `Attribute`, `ontology=`), matching the v1.2.x rename (commit 363a53d):

- **Docs + README** (8 files): all examples, type hints, and prose renamed; `docs/graph-schema.md` filename (and its published URL) intentionally unchanged; api-reference `## Schema` section renamed to `## Ontology`
- **Examples 02/03/06**: renamed + imports re-sorted; all three compile-checked. **`08_ontology_lifecycle.py` intentionally untouched** — its step 8 demonstrates that the legacy names still work
- **Alignment fix found during verification**: api-reference documented `required: bool = False` on `Attribute`, but that field does not exist in `core/models.py` — removed
- Legitimate uses of "schema" left alone (JSON schema in benchmark.md, DB schema prompt in retrieval.md, `SchemaExtensionProposal` which is a real class)

Every renamed reference was verified against `graphrag_sdk/src` on main: `GraphRAG(ontology=...)` + deprecated `schema=` shim, public attribute `self.ontology`, `extract(..., ontology, ...)`, `ontology.entities` precedence, model fields of `Entity`/`Relation`/`Attribute`/`Ontology`, the 11 default entity types list, and the `__init__` exports (old names exist only in `_LEGACY_MODEL_ALIASES`).

Ruff on the touched example files: 5 findings on this branch vs 6 on main — all pre-existing, none introduced.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated guides and API reference to use ontology-based configuration (`Ontology`, `Entity`, `Relation`, `Attribute`) instead of the legacy schema terminology.
  * Revised extraction, ingestion, and strategy documentation to accept and prioritize `ontology` inputs (including “open schema mode” using an empty `Ontology()`).
  * Refreshed all related examples to construct an `Ontology` and pass it into `GraphRAG` via the `ontology=` argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,20 +81,20 @@ async def main():
 asyncio.run(main())
 ```
 
-### 3. Define a schema (optional)
+### 3. Define an ontology (optional)
 
 ```python
-from graphrag_sdk import GraphSchema, EntityType, RelationType
+from graphrag_sdk import Ontology, Entity, Relation
 
-schema = GraphSchema(
+ontology = Ontology(
     entities=[
-        EntityType(label="Person", description="A human being"),
-        EntityType(label="Organization", description="A company or institution"),
-        EntityType(label="Location", description="A geographic location"),
+        Entity(label="Person", description="A human being"),
+        Entity(label="Organization", description="A company or institution"),
+        Entity(label="Location", description="A geographic location"),
     ],
     relations=[
-        RelationType(label="WORKS_AT", description="Is employed by", patterns=[("Person", "Organization")]),
-        RelationType(label="LOCATED_IN", description="Is situated in", patterns=[("Organization", "Location")]),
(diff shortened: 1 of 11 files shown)


Anchel123:
Rebased onto current `main` and addressed the four review comments in https://github.com/FalkorDB/GraphRAG-SDK/pull/308. Leaving this PR open for reference.

PR #288, opened 2026-07-28
feat: chunk-level extraction cache for update() (cache_unchanged_chunks)

Naseem77:
## What

Two changes, both aimed at making `update()` cheap and reliable.

**1. Chunk-level extraction cache** (opt-in: `cache_unchanged_chunks=True`)

A chunk whose text is byte-identical (SHA-256) to one already stored for the same document **skips LLM extraction entirely**. Its entities, relationships and mentions are rebuilt from the live graph and remapped onto the new chunk uids.

Editing one paragraph of a 50-chunk document now costs ~1 extraction instead of 50.

This extends the existing document-level no-op short-circuit down to chunk granularity, reusing the `extractor=` hook `update()` already exposes. No new `update()` function — the existing one gains a flag.

**2. Fix: config validation rejected provider-prefixed model names**

`EmbedderConfig.to_embedder()` builds `azure/text-embedding-3-large` when a provider is set, but the name stored on the graph is unprefixed. `_validate_graph_config` compared the two raw strings — so pointing an existing graph at a provider (same model, same dimensions) raised `ConfigError` and broke **every** update from then on.

Both sides are now normalized before comparing: a leading `<provider>/` is stripped when the prefix is a known litellm provider. The check itself stays — a genuine model swap silently corrupts vector search, since old and new embeddings are never comparable and nothing ever raises.

## How

- **`CachedChunkExtraction`** (new, exported) — a decorator `ExtractionStrategy`. The graph is the cache; nothing extra is stored.
- **3 new `GraphStore` read accessors** + `ChunkEntityRow` / `ChunkRelationshipRow` — schema coupling stays in the storage layer.
- **`cache_unchanged_chunks: bool = False`** on `update()`, `update_sync()`, `apply_changes()`, `apply_changes_sync()`. Stats land in `UpdateResult.metadata["cache_stats"]`.
- **`_bare_model_name()`** — provider list read from litellm once and cached, with a static fallback.

## Safety

- Cache reads happen in Phase 3, while the old chunks still exist (they're deleted at the Phase 5 cutover). The crash-safe pending/commit/rollforward state machine is untouched.
- Deterministic entity ids MERGE onto existing nodes (`SET n +=` preserves embeddings). `source_chunk_ids` remap: own old ids → new uid(s) or dropped; other documents' ids pass through. RELATES provenance is UNIONed.
- Re-emitted mentions keep entities alive through scoped orphan cleanup. Duplicate identical chunks each get their own provenance. Label-less entities skip the node write (MERGE would mint a duplicate) but keep the mention.
- **Fail-open**: any cache lookup or rebuild error falls back to full extraction. Worst case is LLM spend, never data loss.

## Caveats (also on the docstring + CHANGELOG)

- The graph is the cache, so manually deleted entities are **not** resurrected from unchanged chunks.
- Ontology / prompt / model changes do **not** re-extract cached chunks.
- Defaults to `False`; existing behavior is byte-identical until you opt in.

## Testing

**1056 passed, 0 failed**; ruff clean.

- 22 unit tests for the cache: split / remap / merge / fallback matrix, plus the new accessors.
- 4 real-FalkorDB integration tests with a **strict** scripted LLM — extraction of an unchanged chunk raises, so a passing test *is* the proof that caching skipped it.
- 8 new real-FalkorDB edge tests. The key one runs the same edit twice, cached and uncached, and asserts the two graphs come out identical — that states the correctness property directly instead of hand-writing expectations, and it catches faults in the rebuild Cypher that fixture-backed tests can't see. The rest cover reordered chunks, duplicate identical chunks, cross-document provenance isolation, orphan cleanup, stability across repeated updates, and the no-op short circuit.
- 12 tests for the validation fix, covering both directions: the provider-prefix case now passes, a real model change still raises.

Verified end to end against a live server + FalkorDB over the real HTTP path: **18/19 chunks reused (95%)** on a one-paragraph edit, with the changed chunk correctly re-extracted.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in cache for unchanged document chunks during update operations, reducing unnecessary extraction work.
  - Cache statistics now report reused and newly extracted chunks across update workflows.
  - Added the reusable `CachedChunkExtraction` strategy.
  - Improved compatibility with provider-prefixed embedding model names.
- **Documentation**
  - Documented cache configuration, reported statistics, and supporting graph-store accessors.

### DIFF
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -10,6 +10,7 @@ import json
 import logging
 import os
 import re
+from functools import lru_cache
 from typing import Any, Literal, overload
 from uuid import uuid4
 
@@ -145,6 +146,73 @@ _CONTEXT_CLOSE_RE = re.compile(r"</\s*context\s*>", re.IGNORECASE)
 def _neutralize_context_close_tag(text: str) -> str:
     """Disarm any literal ``</context>`` that appears inside untrusted text."""
     return _CONTEXT_CLOSE_RE.sub("</ context>", text)
+
+
+# Fallback provider prefixes for :py:func:`_bare_model_name`, used when
+# litellm isn't installed (it is an optional extra). Covers the providers
+# that actually serve embedding models; the full list is pulled from
+# litellm at runtime when available.
+_FALLBACK_LITELLM_PROVIDERS = frozenset(
+    {
+        "azure",
+        "azure_ai",
+        "bedrock",
+        "cohere",
+        "databricks",
+        "deepinfra",
+        "gemini",
+        "huggingface",
+        "mistral",
+        "nvidia_nim",
+        "ollama",
+        "openai",
+        "openrouter",
+        "together_ai",
+        "vertex_ai",
+        "voyage",
+        "watsonx",
+        "xinference",
+    }
+)
+
+
+@lru_cache(maxsize=1)
+def _litellm_provider_prefixes() -> frozenset[str]:
+    """Known litellm routing prefixes, lowercased.
+
+    Prefers litellm's own ``provider_list`` so the set stays correct as
+    litellm adds providers; falls back to a static set when litellm isn't
+    installed. Cached — the list is static for the process lifetime.
+    """
+    try:
+        import litellm
+
+        providers = {str(getattr(p, "value", p)).lower() for p in litellm.provider_list}
+        if providers:
+            return frozenset(providers | _FALLBACK_LITELLM_PROVIDERS)
+    except Exception:  # pragma: no cover - depends on optional extra
+        logger.debug("litellm provider_list unavailable", exc_info=True)
+    return _FALLBACK_LITELLM_PROVIDERS
+
+
+def _bare_model_name(name: str) -> str:
+    """Strip litellm's provider routing prefix from a model identifier.
+
+    ``"azure/text-embedding-3-large"`` and ``"text-embedding-3-large"`` name
+    the *same* model reached through different endpoints — the prefix tells
+    litellm where to send the request, it is not part of the model identity.
+    Comparing the raw strings therefore reports a mismatch for a graph that
+    merely changed provider, even though the vectors are unchanged.
+
+    Only the first segment is removed, and only when it is a recognised
+    provider, so multi-segment identifiers such as
+    ``"openrouter/anthropic/claude-3"`` keep the part that distinguishes the
+    model. Names without a known prefix are returned unchanged.
+    """
+    head, sep, tail = name.partition("/")
+    if sep and tail and head.lower() in _litellm_provider_prefixes():
+        return tail
+    return name
 
 
 def _strip_and_load_json(text: str) -> Any:
@@ -2878,7 +2946,9 @@ class GraphRAG:
                 stored_dim = result.result_set[0][1]
                 current_model = self.embedder.model_name
 
-                if stored_model and stored_model != current_model:
+                if stored_model and _bare_model_name(stored_model) != _bare_model_name(
+                    current_model
+                ):
                     raise ConfigError(
                         f"Embedding model mismatch: graph was built with "
                         f"'{stored_model}' but current embedder is "
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -150,37 +150,41 @@ def _neutralize_context_close_tag(text: str) -> str:
 def _same_embedding_model(stored: str, current: str) -> bool:
     """Whether two model identifiers name the same embedding model.
 
-    ``"azure/text-embedding-3-large"`` and ``"text-embedding-3-large"`` are the
-    *same* model reached through different endpoints — the leading segment is a
-    routing prefix telling the client where to send the request, it is not part
-    of the model identity. A plain string comparison would reject a graph whose
-    provider changed, even though the stored vectors are unchanged.
+    A leading segment can mean two different things, and the string alone
+    doesn't say which:
 
-    One leading route segment is therefore optional on either side: the names
-    match if they are equal with or without it, which covers a prefix appearing,
-    disappearing, or changing (``"azure/…"`` -> ``"openai/…"``). Comparison is
-    case-insensitive and ignores surrounding whitespace.
+    - a *route* — ``"azure/text-embedding-3-large"`` is the same model as
+      ``"text-embedding-3-large"``, just reached through a different endpoint;
+    - an *owner* — ``"BAAI/bge-m3"`` and ``"ollama/bge-m3"`` are a fp32 release
+      and a quantized build, same dimensions but different vectors.
 
-    Only a whole ``/``-delimited segment can be ignored, never a substring: a
-    looser check would accept ``"text-embedding-3-large"`` against
-    ``"text-embedding-3-large-v2"`` — different models with different vectors —
-    and silently pass the mismatch this guard exists to catch.
+    A bare name is the tell: it carries no route, so a segment present on only
+    one side is one. That case is ignored. When both sides carry a segment they
+    name owners and are compared in full, which keeps the pairs above distinct.
 
-    A routing prefix is indistinguishable from a vendor namespace, so
-    ``"my-org/custom-embedder"`` and ``"custom-embedder"`` also compare equal.
+    This matters because the dimension check can't separate them — every
+    realistic collision here has identical dimensions — and the resulting
+    failure is silent: retrieval still returns results, ranked against vectors
+    from another model.
+
+    Comparison is case-insensitive and ignores surrounding whitespace. Only a
+    whole ``/`` segment is ever ignored, never a substring, so
+    ``"text-embedding-3-large"`` and ``"text-embedding-3-large-v2"`` stay
+    distinct.
     """
-
-    def _variants(name: str) -> set[str]:
-        name = name.strip().lower()
-        if not name:
-            return set()
-        forms = {name}
-        _, sep, tail = name.partition("/")
-        if sep and tail:
-            forms.add(tail)
-        return forms
-
-    return bool(_variants(stored) & _variants(current))
+    a = stored.strip().lower()
+    b = current.strip().lower()
+    if not a or not b:
+        return a == b
+    if a == b:
+        return True
+    _, a_sep, a_tail = a.partition("/")
+    _, b_sep, b_tail = b.partition("/")
+    if a_sep and a_tail and not b_sep:
+        return a_tail == b
+    if b_sep and b_tail and not a_sep:
+        return b_tail == a
+    return False
 
 
 def _strip_and_load_json(text: str) -> Any:
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -54,6 +54,9 @@ from graphrag_sdk.ingestion.chunking_strategies.sentence_token_cap import (
     SentenceTokenCapChunking,
 )
 from graphrag_sdk.ingestion.extraction_strategies.base import ExtractionStrategy
+from graphrag_sdk.ingestion.extraction_strategies.cached_chunk_extraction import (
+    CachedChunkExtraction,
+)
 from graphrag_sdk.ingestion.extraction_strategies.entity_extractors import (
     DEFAULT_ENTITY_TYPES,
     _strip_markdown_fences,
@@ -1923,6 +1926,7 @@ class GraphRAG:
         chunker: ChunkingStrategy | None = None,
         extractor: ExtractionStrategy | None = None,
         resolver: ResolutionStrategy | None = None,
+        cache_unchanged_chunks: bool = False,
         if_missing: Literal["error", "ingest"] = "error",
         ctx: Context | None = None,
     ) -> UpdateResult:
@@ -1984,6 +1988,23 @@ class GraphRAG:
                 with no extra plumbing. Required in text mode.
             loader / chunker / extractor / resolver: Per-call strategy
                 overrides, identical to ``ingest()``.
+            cache_unchanged_chunks: When ``True``, wrap the extractor in
+                :class:`~graphrag_sdk.ingestion.extraction_strategies.cached_chunk_extraction.CachedChunkExtraction`:
+                new chunks whose text is byte-identical to an existing
+                chunk of this document skip LLM extraction — their
+                entities/relations/mentions are rebuilt from the live
+                graph and remapped onto the new chunk uids. Only changed
+                chunks pay for extraction. Cache effectiveness is
+                reported in ``UpdateResult.metadata["cache_stats"]``
+                (``cached_chunks`` / ``extracted_chunks``). Two semantic
+                caveats: (1) the graph is the cache, so manually deleted
+                entities are NOT resurrected from unchanged chunks; (2)
+                if the ontology, prompts, or model changed since ingest,
+                unchanged chunks keep their previously extracted data.
+                Leave ``False`` (the default) to force full re-extraction
+                in those situations. No effect on the
+                ``if_missing="ingest"`` fresh-ingest fallthrough (a new
+                document has no chunks to reuse).
             if_missing: ``"error"`` (default) raises ``DocumentNotFoundError``
                 when the id is unknown. ``"ingest"`` falls through to
                 ``ingest()`` for upsert semantics.
@@ -2109,10 +2130,24 @@ class GraphRAG:
         # Contradictions surface here rather than mid-extraction.
         await self._ensure_ontology_initialized()
 
+        # Chunk-level extraction cache (opt-in). Wraps the effective
+        # extractor so byte-identical chunks are rebuilt from the live
+        # graph instead of re-extracted. Must read the OLD chunks, which
+        # still exist until the Phase 5 cutover — safe by construction.
+        active_extractor = extractor or self._default_extractor()
+        cache_wrapper: CachedChunkExtraction | None = None
+        if cache_unchanged_chunks:
+            cache_wrapper = CachedChunkExtraction(
+                inner=active_extractor,
+                graph_store=self._graph_store,
+                document_id=resolved_id,
+            )
+            active_extractor = cache_wrapper
+
         pipeline = IngestionPipeline(
             loader=loader or TextLoader(),  # unused (text is provided below)
             chunker=chunker or SentenceTokenCapChunking(),
-            extractor=extractor or self._default_extractor(),
+            extractor=active_extractor,
             resolver=resolver or ExactMatchResolution(),
             graph_store=self._graph_store,
             vector_store=self._vector_store,
@@ -2190,12 +2225,19 @@ class GraphRAG:
             f"wrote {pipeline_result.chunks_indexed} new chunks"
         )
 
+        result_metadata = dict(pipeline_result.metadata)
+        if cache_wrapper is not None:
+            result_metadata["cache_stats"] = {
+                "cached_chunks": cache_wrapper.cached_chunk_count,
+                "extracted_chunks": cache_wrapper.extracted_chunk_count,
+            }
+
         return UpdateResult(
             document_info=DocumentInfo(uid=resolved_id, path=doc_path, metadata=loaded_metadata),
             nodes_created=pipeline_result.nodes_created,
             relationships_created=pipeline_result.relationships_created,
             chunks_indexed=pipeline_result.chunks_indexed,
-            metadata=pipeline_result.metadata,
+            metadata=result_metadata,
             chunks_deleted=chunks_deleted,
             entities_deleted=entities_deleted,
             replaced_existing=True,
@@ -2324,6 +2366,7 @@ class GraphRAG:
         chunker: ChunkingStrategy | None = None,
         extractor: ExtractionStrategy | None = None,
         resolver: ResolutionStrategy | None = None,
+        cache_unchanged_chunks: bool = False,
         max_concurrency: int = 3,
         update_concurrency: int = 1,
         ctx: Context | None = None,
@@ -2377,6 +2420,11 @@ class GraphRAG:
                 ``added``/``modified``. ``deleted`` ignores this.
             resolver: Override the resolution strategy for ``added``/
                 ``modified``. ``deleted`` ignores this.
+            cache_unchanged_chunks: Forwarded to ``update()`` for the
+                ``modified`` list — unchanged chunks skip LLM extraction
+                and are rebuilt from the live graph. See
+                :meth:`update` for semantics and caveats. ``added`` and
+                ``deleted`` ignore this.
             max_concurrency: Parallelism cap for ``ingest()`` of the
                 ``added`` list. Matches ``ingest()``'s own knob and the
                 ``add`` step is pure ingestion with no orphan-cleanup
@@ -2469,6 +2517,7 @@ class GraphRAG:
                             chunker=chunker,
                             extractor=extractor,
                             resolver=resolver,
+                            cache_unchanged_chunks=cache_unchanged_chunks,
                             if_missing="ingest",
                             ctx=ctx.child(),
                         )
@@ -3138,6 +3187,7 @@ class GraphRAG:
         chunker: ChunkingStrategy | None = None,
         extractor: ExtractionStrategy | None = None,
         resolver: ResolutionStrategy | None = None,
+        cache_unchanged_chunks: bool = False,
         if_missing: Literal["error", "ingest"] = "error",
         ctx: Context | None = None,
     ) -> UpdateResult:
@@ -3161,6 +3211,7 @@ class GraphRAG:
                 chunker=chunker,
                 extractor=extractor,
                 resolver=resolver,
+                cache_unchanged_chunks=cache_unchanged_chunks,
                 if_missing=if_missing,
                 ctx=ctx,
             )
@@ -3193,6 +3244,7 @@ class GraphRAG:
         chunker: ChunkingStrategy | None = None,
         extractor: ExtractionStrategy | None = None,
         resolver: ResolutionStrategy | None = None,
+        cache_unchanged_chunks: bool = False,
         max_concurrency: int = 3,
         update_concurrency: int = 1,
         ctx: Context | None = None,
@@ -3215,6 +3267,7 @@ class GraphRAG:
                 chunker=chunker,
                 extractor=extractor,
                 resolver=resolver,
+                cache_unchanged_chunks=cache_unchanged_chunks,
                 max_concurrency=max_concurrency,
                 update_concurrency=update_concurrency,
                 ctx=ctx,
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -10,7 +10,6 @@ import json
 import logging
 import os
 import re
-from functools import lru_cache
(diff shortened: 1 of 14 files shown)


galshubeli:
Ran the suite on this branch: 166 passed, 12 skipped. Design holds up, and confirmed there's no duplicate update path — `update()` just gained a flag.

**1. `CachedChunkExtraction` is exported but only safe inside `update()` — Medium**

It never checks that the doc id it reads differs from the one the pipeline is writing chunks to. That only holds today because `update()` writes under a pending node while the cache reads `resolved_id`. If they ever coincide (wired into `ingest()`, or the pending indirection changes), `get_document_chunk_texts()` returns old *and* new rows with identical text, `setdefault` picks whichever came back first, and a chunk can map to itself → no MENTIONED_IN rows → counted as cached, skips the extractor, entities silently dropped. Fail-open can't catch it (no exception) and `cache_stats` still looks fine. The unit tests miss it because `FakeGraphStore` returns an old-only list.

Fix is ~3 lines: skip any returned id in `{c.uid for c in chunks.chunks}`.

**2. `get_relationships_for_chunks` scans every RELATES edge — Medium (perf)**
`any(c IN r.source_chunk_ids WHERE c IN $cids)` isn't index-servable, so cost scales with total graph size, not document size. Won't show at 19 chunks. Start the scan from the entity set you already have.

**3. Duplicate accessor — Low**
`get_document_chunk_texts` and `get_document_chunk_ids` are the same traversal, one extra column, and both run in the same `update()` call. Pass the existing `old_chunk_ids` snapshot in instead.

Approve once #1 is addressed; #2/#3 are fine as follow-ups.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -144,6 +148,73 @@ def _neutralize_context_close_tag(text: str) -> str:
     return _CONTEXT_CLOSE_RE.sub("</ context>", text)
 
 
+# Fallback provider prefixes for :py:func:`_bare_model_name`, used when
```
galshubeli: why we need this?

Naseem77: @galshubeli we check if user has litellm if so we take the provider names from there and strip the prefix, if they dont have litellm we use the fallback which is the hardcoded list to stripe the prefix, the reason we have a litellm list because that it can have provider names our list doesn't have.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -144,6 +147,42 @@ def _neutralize_context_close_tag(text: str) -> str:
...
+        name = name.strip().lower()
+        if not name:
+            return set()
+        forms = {name}
+        _, sep, tail = name.partition("/")
+        if sep and tail:
+            forms.add(tail)
+        return forms
+
+    return bool(_variants(stored) & _variants(current))
```
galshubeli: Dropping the hardcoded provider list was the right call, but the replacement is **strictly more permissive** than what it replaced. The old version only stripped a segment when it was a known provider; this strips any first segment, so every pair the old code rejected still passes, plus a lot more.

The rule reduces to "do the model basenames match", and that collides badly with HuggingFace-style repo ids. I ran the new function:

| stored | current | result |
|---|---|---|
| `sentence-transformers/all-MiniLM-L6-v2` | `myorg/all-MiniLM-L6-v2` | **match** |
| `BAAI/bge-m3` | `ollama/bge-m3` | **match** |
| `intfloat/e5-large-v2` | `rando/e5-large-v2` | **match** |
| `openai/text-embedding-3-large` | `mistralai/text-embedding-3-large` | **match** |

These aren't hypothetical. `examples/04_custom_provider.py` points people directly at `SentenceTransformer.encode`, and a custom `Embedder` returns whatever `model_name` it likes — which for sentence-transformers is an `org/model` repo id. Row 1 is a finetune of a base model: same architecture, same dimensions, completely different vectors. Row 2 is a quantized GGUF against fp32 weights, also same dimensions.

Every one of these silently passes the guard whose entire job is catching them.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -144,6 +147,42 @@ def _neutralize_context_close_tag(text: str) -> str:
...
+    match if they are equal with or without it, which covers a prefix appearing,
+    disappearing, or changing (``"azure/…"`` -> ``"openai/…"``). Comparison is
+    case-insensitive and ignores surrounding whitespace.
+
+    Only a whole ``/``-delimited segment can be ignored, never a substring: a
+    looser check would accept ``"text-embedding-3-large"`` against
+    ``"text-embedding-3-large-v2"`` — different models with different vectors —
+    and silently pass the mismatch this guard exists to catch.
+
+    A routing prefix is indistinguishable from a vendor namespace, so
```
galshubeli: This line is the strongest argument against solving this with strings at all:

> A routing prefix is indistinguishable from a vendor namespace

That's correct, and it's not a limitation of *this* implementation — no string rule can tell `azure/x` (routing, ignorable) from `BAAI/x` (identity, load-bearing), because the information isn't in the string. So the choice is which way to be wrong, and the PR picks the silent direction.

The two failure modes aren't symmetric:

- **False positive** (raise on a compatible pair): loud, immediate, user can act on it. This is the bug being fixed.
- **False negative** (pass on an incompatible pair): silent, degrades retrieval invisibly, surfaces months later as "search got worse" — if ever.

Trading the first away to buy the second is a bad trade even at a good exchange rate.

**The good news: you already compute the real answer and throw it away.** A few lines below this, `_validate_graph_config` embeds a fixed probe string and uses only `len(probe)`. Store that vector on the config node and compare it:

```python
# _write_graph_config
"SET c.embedding_model = $model, "
"c.embedding_dimension = $dim, "
"c.embedding_probe = $probe, "          # vector for "dim_check"

# _validate_graph_config
sim = cosine(probe, stored_probe)
if sim < 0.99:
    raise ConfigError(
        f"Embedding function mismatch: graph built with '{stored_model}', "
        f"current embedder '{current_model}' produces cosine {sim:.4f} on a "
        f"fixed probe. These embeddings are not comparable."
    )
```

- Costs nothing — the probe call already happens on every validation.
- Measures vector comparability directly instead of inferring it from a name.
- **Passes** the Azure/OpenAI route change this PR is about (cosine ~1.0).
- **Catches** all four rows in my table above, plus `dimensions=` changes and silent vendor model refreshes — none of which any naming rule can see.
- No vendor table, no litellm import, no provider-specific behavior.

Cosine with a threshold rather than exact equality, since batching/fp16/hardware make bit-identical output unrealistic. Quantized-vs-fp32 of the same model lands ~0.98-0.995, so ~0.99 is the interesting boundary and should be tunable. First 64 dims are plenty if you'd rather not store a full vector. Graphs with no stored probe fall back to today's behavior.

The model name then becomes a hint: mismatch logs a warning, the vector decides.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -2829,7 +2914,7 @@ async def _validate_graph_config(self, *, ctx: Context | None = None) -> None:
                 stored_dim = result.result_set[0][1]
                 current_model = self.embedder.model_name
 
-                if stored_model and stored_model != current_model:
+                if stored_model and not _same_embedding_model(stored_model, current_model):
```
galshubeli, line 2921: Worth stepping back on this comparison generally.

Inside the SDK both sides are already symmetric — `_write_graph_config` writes `self.embedder.model_name` and this reads it back to compare against `self.embedder.model_name`. A graph written by this SDK and read by this SDK cannot hit the prefix mismatch. The asymmetry is entirely the server's `EmbedderConfig.to_embedder()` storing a bare name and later constructing a prefixed one.

So a global invariant is being loosened for every SDK user to compensate for a caller bug in a different repo. And because `_write_graph_config` uses `MERGE ... SET`, affected graphs rewrite the stored name on their next successful ingest — only the bootstrap needs a one-shot fix.

The original incident (a `ConfigError` that then broke every subsequent update) is a legitimate complaint, but it's a complaint about an **un-overridable, inscrutable error**, not about the error existing. That's fixable without widening the check:

```
if raw names match                     -> pass
if probe vectors are similar           -> pass, rewrite stored name
if basenames match and dims match      -> ConfigError naming the override flag
otherwise                              -> ConfigError
```

The third branch is where this PR's users land, and the message carries its own fix: *"same base model via a different provider — pass `embedding_model_policy='allow_provider_change'` if these vectors are comparable."* One deliberate line during a rare, supervised migration, and it self-heals on the next write. That's a much better shape than permanently relaxing the check for everyone.

Preference order: fix `EmbedderConfig` upstream > probe-vector fingerprint > explicit override flag > the current string widening.

graphrag_sdk/tests/test_facade.py
```diff
@@ -1174,6 +1174,153 @@ async def test_no_config_node_passes(self, mock_conn, embedder):
...
+        """The guard must keep catching a real swap — a prefix must not be
+        able to mask two different models."""
+        with pytest.raises(ConfigError, match="Embedding model mismatch"):
+            await self._retrieve_with(
+                mock_conn,
+                self._embedder_named("azure/text-embedding-ada-002"),
+                "azure/text-embedding-3-large",
+            )
+
+    async def test_same_model_under_different_vendors_passes(self, mock_conn):
```
galshubeli: This is the part I'd push back on hardest: a test that asserted a mismatch **raises** has been inverted to assert it **passes**. `test_unknown_prefix_still_raises` was correct — `team-a/mock-embedder` and `team-b/mock-embedder` are two different models.

The justification given doesn't hold:

> The dimension check below still guards the case that actually corrupts retrieval.

It doesn't. The dimension check is `stored_dim != self._embedding_dimension`, and every realistic instance of this collision has **identical dimensions**:

- `sentence-transformers/all-MiniLM-L6-v2` vs a finetune under another org — both 384
- `BAAI/bge-m3` (fp32) vs `ollama/bge-m3` (quantized) — both 1024
- `intfloat/e5-large-v2` vs a finetune — both 1024

Same dims, different vectors, guard passes, cosine similarity against the stored embeddings is meaningless. That is precisely "the case that actually corrupts retrieval", and it is the one case neither check now catches.

Retrieval also fails *soft* here — you get results back, ranked badly. There's no error and no obvious symptom, so this can sit in a production graph indefinitely.

I'd restore this test. If the vendor-namespace collision has to be accepted as a deliberate tradeoff, it needs to be stated as a known false-negative with the corruption risk spelled out, not as a case the dimension check covers.

graphrag_sdk/tests/test_facade.py
```diff
@@ -1174,6 +1174,153 @@ async def test_no_config_node_passes(self, mock_conn, embedder):
...
+            ("vertex_ai/textembedding-gecko", "textembedding-gecko"),
+            # Identical on both sides, with and without a route.
+            ("text-embedding-3-large", "text-embedding-3-large"),
+            ("azure/text-embedding-3-large", "azure/text-embedding-3-large"),
+            # Case and surrounding whitespace are not part of the identity.
+            ("Text-Embedding-3-Large", "text-embedding-3-large"),
+            ("  azure/text-embedding-3-large  ", "text-embedding-3-large"),
+            # Multi-segment routes collapse to the trailing model id.
+            ("openrouter/anthropic/claude-3", "anthropic/claude-3"),
+            # Any vendor namespace, not just the well-known providers.
```
galshubeli: This parameter encodes the widening as intended behavior, so it's worth being explicit that `my-org/custom-embedder` == `custom-embedder` also implies `my-org/custom-embedder` == `other-org/custom-embedder`.

That second form is the one that bites, and it's reachable through the documented custom-embedder path in `examples/04_custom_provider.py`.

graphrag_sdk/src/graphrag_sdk/storage/graph_store.py
```diff
@@ -680,6 +686,135 @@ async def get_document_chunk_ids(self, document_id: str) -> list[str]:
...
+        once per chunk id. The chunk-id intersection is computed client-side
+        from the returned provenance list.
+        """
+        rows: list[ChunkRelationshipRow] = []
+        for start in range(0, len(chunk_ids), self._BATCH_SIZE):
+            batch = chunk_ids[start : start + self._BATCH_SIZE]
+            batch_set = set(batch)
+            result = await self._conn.query(
+                "MATCH (a:__Entity__)-[r:RELATES]->(b:__Entity__) "
+                "WHERE any(c IN r.source_chunk_ids WHERE c IN $cids) "
```
galshubeli, line 788: This is a full graph edge scan sitting in the hot path of the feature.

`source_chunk_ids` is a plain list property, so nothing can index this predicate — every `RELATES` edge in the database is visited, once per 500-chunk batch. The docstring frames "the edge scan runs ONCE per batch, not once per chunk id" as the optimization, but the baseline it optimizes from is O(all edges in the graph).

So cost scales with total graph size, not with the document being updated. Editing one paragraph of a small document in a graph holding a few hundred other documents pays for all of them — which is the exact cost profile `cache_unchanged_chunks` exists to remove. The gain gets thinner as the graph grows, which is backwards.

Chunk ids are indexed, so reaching the edges through the chunks avoids the scan:

```cypher
MATCH (c:Chunk) WHERE c.id IN $cids
MATCH (e:__Entity__)-[:MENTIONED_IN]->(c)
MATCH (e)-[r:RELATES]-(o:__Entity__)
RETURN ...
```

The provenance intersection is already computed client-side from the returned `source_chunk_ids`, so the filtering below doesn't need to change.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/cached_chunk_extraction.py
```diff
@@ -0,0 +1,325 @@
...
+        )
+
+    async def extract(self, chunks: TextChunks, ontology: Ontology, ctx: Context) -> GraphData:
+        """Split chunks into cached vs. new, rebuild the former from the
+        graph, extract the latter with the inner strategy, and merge."""
+        try:
+            old_by_hash, all_old_ids = await self._old_chunks_by_hash(
+                {c.uid for c in chunks.chunks}
+            )
+        except Exception as exc:
```
galshubeli, line 396: Fail-open is right for a cache miss but wrong for an infrastructure error.

The realistic trigger for this `except` is a `DatabaseError` — dropped connection, query timeout — not a logic fault. When the graph is unreachable, this decides to run full LLM extraction over every chunk in the document. That extraction *succeeds* (it only needs the LLM), and then the write phase hits the same dead connection and fails anyway.

So an error that could surface in milliseconds instead costs a full document of LLM spend and lands in the same place. The PR frames the worst case as "LLM spend, never data loss" — this is that worst case being reached by the *most likely* failure mode rather than an unlikely one.

Same at the rebuild fallback below (line 311).

Letting `DatabaseError` and `LatencyBudgetExceededError` propagate and keeping fail-open for everything else would fix it. `_validate_graph_config` already uses exactly this pattern — `except ConfigError: raise` / `except LatencyBudgetExceededError: raise` / `except Exception: log`.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/cached_chunk_extraction.py
```diff
@@ -0,0 +1,325 @@
...
+        self._graph_store = graph_store
+        self._document_id = document_id
+        # Filled during extract() so the caller can report cache stats.
+        self.cached_chunk_count = 0
+        self.extracted_chunk_count = 0
+
+    async def _old_chunks_by_hash(self, exclude: set[str]) -> tuple[dict[str, str], set[str]]:
+        """Return (sha256(chunk text) -> old chunk id, all old chunk ids).
+
+        Ids in ``exclude`` — the uids of the chunks currently being written —
```
galshubeli, line 118: This docstring describes a silent-data-loss path that the class can't structurally prevent, and the class is public API (exported from both `graphrag_sdk.__init__` and `ingestion.__init__`).

The failure is spelled out here in full: a chunk hashes to itself, the rebuild finds no mentions, emits empty `GraphData`, skips the extractor, drops the entities — **with no exception for the fail-open path to catch** — and `cached_chunk_count` still reports a hit. Every safety net in the design misses this one.

The `exclude` set only helps when the new chunks carry different uids from the old. That's a property of how `update()` happens to wire this up, not an invariant of the class. Someone constructing `CachedChunkExtraction(inner=..., graph_store=..., document_id=<live doc>)` and running it over that document's own chunks gets entities deleted and a success result.

The most recent commit before the embedding work is `fix: never treat a chunk as its own cache entry`, which is decent evidence this is reachable in practice.

Either keep it internal for now (drop it from the public exports) and revisit once the contract has settled, or make the requirement enforceable — e.g. take the read document id and the write document id as separate arguments and raise when they're equal, so the caller has to state intent rather than get it right by accident.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/cached_chunk_extraction.py
```diff
@@ -0,0 +1,325 @@
...
+
+        nodes: list[GraphNode] = []
+        for eid, props in ent_props.items():
+            label = ent_labels[eid]
+            if not label:
+                # No usable label: upsert_nodes would MERGE on a fallback
+                # label and mint a duplicate node instead of matching this
+                # one. Skip the node write — the mention above still keeps
+                # the entity alive through orphan cleanup.
+                logger.warning("Skipping cache node write for label-less entity %s", eid)
```
galshubeli: This drop isn't visible in `cache_stats`, which makes the feature's only operational signal unreliable.

A label-less entity is skipped with a log line. Nothing raises, so fail-open never fires, and the chunk is still counted in `cached_chunk_count` as fully reused. Same for rows discarded by the `not cid or not eid` guard in `get_entities_mentioned_in_chunks`.

So the graph can come out thinner than a full re-extraction would produce while `UpdateResult.metadata["cache_stats"]` reports a clean 100% hit rate. For a fail-open cache that number is how an operator knows things are healthy — it shouldn't be the one thing blind to partial rebuilds. A `skipped_entities` counter alongside `cached_chunks` / `extracted_chunks` would make it observable without changing behavior.

graphrag_sdk/src/graphrag_sdk/storage/graph_store.py
```diff
@@ -680,6 +686,135 @@ async def get_document_chunk_ids(self, document_id: str) -> list[str]:
...
+            result = await self._conn.query(
+                "UNWIND $cids AS cid "
+                "MATCH (e:__Entity__)-[:MENTIONED_IN]->(c:Chunk {id: cid}) "
+                "RETURN cid, e.id, labels(e), e.name, e.type, e.description, "
+                "e.source_chunk_ids",
+                {"cids": batch},
+            )
+            for row in result.result_set or []:
+                cid, eid, labels, name, etype, description, source_chunk_ids = row
+                if not cid or not eid:
```
galshubeli, line 742: `not cid` is falsy, not absent — a chunk id of `"0"` is discarded here as though the row were malformed.

Almost certainly unreachable with current uid generation, but this is a data-integrity guard, and conflating falsy with missing inside one stops being theoretical the day id generation changes. `if cid is None or eid is None` costs nothing. Same at line 795 for `start_id` / `end_id`.

galshubeli (commented):
Re-reviewed after the embedding-match rework. Splitting this into the two independent changes.

### Cache (the bulk of the PR) — solid

I specifically tried to break several things and couldn't:

- `rel_by_pair` keyed on `(start, end)` ignoring `rel_type` is **correct**, not lossy — `upsert_relationships` does `MERGE (a)-[r:RELATES]->(b)`, so storage already permits one edge per pair. The rebuild matches the schema.
- Provenance remap (own old ids to new uids, other documents' ids passed through) is right, and the `RELATES` asymmetry is safe because `upsert_relationships` UNIONs.
- Duplicate-text chunks via `id_map: old_id -> list[new_uid]` are handled properly.
- `_merge` ordering `[cached, extracted]` with union-on-`source_chunk_ids` and last-wins elsewhere gives fresh extraction precedence. Correct.
- The scalar-vs-list guards in the new accessors are genuinely good defensive work.

Comments below are the real ones: the edge scan in `get_relationships_for_chunks`, fail-open on database errors, and the public export of `CachedChunkExtraction`. None are blocking-by-design, all are worth fixing before merge.

### Embedding model matching — going the wrong way

Dropping the hardcoded provider list was right. But the replacement is **strictly more permissive** than what it replaced: the old rule only stripped known providers, the new one strips any first segment. Everything the old code rejected still passes, plus more.

I ran the new function against realistic names:

| stored | current | result |
|---|---|---|
| `sentence-transformers/all-MiniLM-L6-v2` | `myorg/all-MiniLM-L6-v2` | **match** |
| `BAAI/bge-m3` | `ollama/bge-m3` | **match** |
| `intfloat/e5-large-v2` | `rando/e5-large-v2` | **match** |

All same-dimension, all different vectors, all silently accepted. These are reachable through the custom-embedder path the SDK documents in `examples/04_custom_provider.py`, which points at `SentenceTransformer` — whose model names are exactly `org/model` repo ids.

`test_unknown_prefix_still_raises` was inverted to assert these now pass, on the grounds that "the dimension check below still guards the case that actually corrupts retrieval." That claim isn't right — the dimension check compares `stored_dim != embedding_dimension`, and every case above has identical dimensions. Retrieval then fails soft: results come back, ranked by cosine against incomparable vectors, with no error.

The docstring's own line — *"a routing prefix is indistinguishable from a vendor namespace"* — is exactly the problem, and it's not fixable in string space, because the information isn't in the string.

**The useful part: you already compute the real answer and discard it.** `_validate_graph_config` embeds a fixed probe string and uses only `len(probe)`. Store that vector on the config node and compare cosine — free (the call already happens), passes the Azure/OpenAI route change this PR is about, and catches all three rows above plus `dimensions=` changes and silent vendor model refreshes. Details on `_same_embedding_model`.

Ranked: fix `EmbedderConfig` upstream (both sides are already symmetric inside the SDK — the asymmetry comes from the server) > probe fingerprint > explicit `embedding_model_policy` override > current string widening.

**Either way this should be a separate PR.** It's unrelated to the cache, it relaxes a global invariant, and in a 2,100-line change where the rest is strong it'll get carried along on that strength.

graphrag_sdk/tests/test_facade.py
Naseem77: @galshubeli you're right, my justification was wrong. Reverted the test, and tightened the rule instead of adding machinery. The checks now:

**Check 1 — embedder name**
`stored_model` (from the graph) vs `self.embedder.model_name` (now), via `_same_embedding_model`. Not `==` — it ignores a route segment, but only when the other side has none. A bare name carries no route, so a segment on one side alone is one; two qualified names are two owners and compare in full.

**Check 2 — stored dimension**
`stored_dim` (from the graph) vs `self._embedding_dimension` (config). Plain `!=`, exact.

**Check 3 — probe length**
Embeds a fixed string once and compares `len(probe)` against `self._embedding_dimension`. Catches a fresh graph where the configured dim never matched the model in the first place.

Separate `if`s, so passing one never skips the next.

Your four rows all raise now:

| stored | current | before | after |
|---|---|---|---|
| `sentence-transformers/all-MiniLM-L6-v2` | `myorg/all-MiniLM-L6-v2` | pass | **raise** |
| `BAAI/bge-m3` | `ollama/bge-m3` | pass | **raise** |
| `intfloat/e5-large-v2` | `rando/e5-large-v2` | pass | **raise** |
| `openai/text-embedding-3-large` | `mistralai/text-embedding-3-large` | pass | **raise** |

The prod case (`mock-embedder` vs `azure/mock-embedder`) still passes.

Cost: `azure/x` -> `openai/x` (route changed while both stay qualified) now raises. Rare next to what it catches, and it fails loud.

On the probe fingerprint — I built it, then threw it away. The 0.99 threshold and the 64-dim truncation are both guesses I can't back with data, and a finetune of a base model could plausibly still clear 0.99 on the first 64 dims, which is the exact case it was meant to catch. It also turns a silent-wrong into a loud-wrong for every existing graph if a vendor re-tunes a model. Worth doing, but only after measuring real base-vs-finetune pairs to pick the threshold. Happy to take it as a follow-up.

Still working through your other comments (the `RELATES` scan, fail-open on `DatabaseError`, the public-export question, `skipped_entities`) — will reply on those threads separately.

galshubeli:
**Blocker (1)**

- **#1 — cached relationships beat freshly extracted ones** (`cached_chunk_extraction.py:295`). `_merge` concatenates edges, the resolver (`exact_match.py:69-84`) keeps the first (cached) one and drops the rest. Corrections get discarded and provenance is lost → edge deleted on a later update. Silent data loss that full re-extraction wouldn't produce; breaks the PR's own "never data loss" claim. Reproduced:

  ```
  after _merge:   2 edges (concatenated, not unioned)
      {'fact': 'Alice works at Acme (OLD/WRONG)',  'source_chunk_ids': ['newA']}
      {'fact': 'Alice is CTO of Acme (CORRECTED)', 'source_chunk_ids': ['newB']}

  after resolver: 1 edge survives
      fact:             Alice works at Acme (OLD/WRONG)   <-- correction discarded
      source_chunk_ids: ['newA']                          <-- newB lost
  ```

  Note the resolver's node path merges properties from duplicates; the relationship path rebuilds from the first occurrence and drops the rest entirely, so there's no partial salvage. `delete_stale_relationships` then empties the list on a later update and `WHERE size(r.source_chunk_ids) = 0 ... DELETE r` removes the edge.

  Fix: union relationships in `_merge` keyed on `(start_node_id, type, end_node_id)`, mirroring the node path — fresh properties win, `source_chunk_ids` unioned. `test_cached_and_uncached_updates_converge` misses this because `C_ALICE` and `C_DAVE` have disjoint entity pairs, so no edge ever lands in both a cached and an extracted part.

**Should fix before merge (2)**

- **#2 — the documented fail-open is defeated by this PR's own `connection.py` change** (`cached_chunk_extraction.py:314`, `:349`). `DatabaseError` now wraps permanent Cypher errors too, so the re-raise turns a recoverable case into a hard failure on a document that `cache_unchanged_chunks=False` handles fine. `_is_non_transient` only matches `already indexed` / `already exists` / `unknown index`, so a rejected predicate takes the retry path and gets wrapped at the exhausted-retries line — burning the full retry budget first. Narrow the re-raise, or drop the fail-open claim from the docstring and CHANGELOG.

- **#3 — endpoint check aborts the whole document's cache in a routine case** (`cached_chunk_extraction.py:227`). `entity_type_map` is built from merged entities across all chunks (`graph_extraction.py:616-623`), so an edge whose endpoint is only mentioned in another chunk is normal output, not corruption. The check treats it as corruption and the whole document falls back to full extraction. Not unsafe, but the feature quietly stops working. Scope the check per old chunk rather than skipping the edge — skipping reintroduces the provenance loss the check was added for.

**Nits (3)**

- **#4 — full RELATES scan per batch** (`graph_store.py:787`). No index-usable predicate, so every cached update scans every `RELATES` edge in the graph, including other documents'. Editing one paragraph of a 5,000-chunk doc is 10 full edge scans. Fine as a follow-up.
- **#5 — stale docstring** (`graph_store.py:724`). Says the consumer "keeps the mention" on `label is None`; it now raises `_UnrebuildableCacheEntry` and falls back.
- **#6 — `ping()` reports a packaging error as "server down"** (`connection.py:260`). `_ensure_client()` raises bare `ImportError` for a missing `falkordb` package, now caught by `except Exception` → `return False`. A health endpoint will point operators at the database instead of a broken install.

Only #1 blocks merge. The embedding-match rework, `DatabaseError` at the connection layer, and the label-less `_UnrebuildableCacheEntry` fix from the last round all look right — #1 just sits underneath them, in the merge path every cached update goes through.

galshubeli (approved):

PR #290, opened 2026-08-06
docs: structured data ingestion design proposal

galshubeli:
Design proposal for structured data ingestion. **Docs only — no SDK code.**

Refs: [FalkorDB/research#82](https://github.com/FalkorDB/research/issues/82) (POC) · design from [research#65](https://github.com/FalkorDB/research/issues/65) · supersedes #74

## The problem

`rag.ingest(...)` assumes prose. Structured inputs break three of its assumptions at once: the schema is already known, row identity is explicit, and typed columns shouldn't be re-discovered by an LLM. Today a mixed corpus needs a bespoke loader per format or pre-flattening to text, which throws away the structure that made the source valuable.

The design has to serve differently-shaped CSVs, nested JSON, a table lifted out of a PDF, and an existing graph that may or may not have its own ontology — all landing in one connected, traversable graph.

## The design

Reduce every structured source to a **stream of flat records**, and let **one declarative mapping** turn records into typed nodes and `RELATES` edges. Three ideas carry it:

1. **One intermediate representation** (`RecordBatch`). A PDF table becomes a record stream carrying the PDF's `DocumentInfo`, so its rows are chunks of the *same* `Document` as the prose. An existing graph becomes two streams — nodes and edges — because a graph is a node table plus an edge table. Neither is a special case any more.
2. **The mapping *is* an ontology fragment.** `mapping.to_ontology()` → validate / merge / bootstrap / reject. That single fact is the whole answer to "the existing graph may or may not have an ontology."
3. **Identity is declared once on the entity type**, not per source (`Entity(identity=["name"])`). A PDF mention of `Acme Corp` and a CSV row with `org_name="Acme Corp"` compute the *same* id and MERGE onto the *same* node — connected at write time, no similarity threshold, no LLM. A deterministic `AliasMatchResolution` bridges types keyed by a real business key (`sku`), since extraction can only ever produce a name.

Supporting decision with the most leverage: **a record is persisted as a `Chunk`** in the normal lexical graph. That inherits `update()`/`delete_document()` orphan cleanup, chunk retrieval over rows, and Zero-Loss provenance for free.

The result: **not one retrieval strategy changes.** Entity search, edge-fact search, chunk search, neighbour expansion and text-to-Cypher already see structured data. The doc includes a table justifying that claim path by path.

## Design review findings

The two central claims were walked against the code before publishing. Both hold, and the walk surfaced three requirements and one gap — all folded into the document:

- Structured writes **must** set `name` on nodes and `fact` **and** `source_chunk_ids` on `RELATES` edges. Omitting any one silently breaks entity embedding, edge embedding, or stale-edge GC respectively.
- **Correctness trap:** record chunk uids must key on the *run's* effective document uid, never the canonical id. `update()` runs against `pending_id = f"{resolved_id}__pending__{uuid4().hex[:8]}"`, and `rollforward_cutover()` deletes the live document's chunks before promoting the pending — so canonical-keyed chunks would MERGE onto the same nodes and the cutover would delete the chunks it is about to promote.
- `get_document_entity_candidates()` has no `LIMIT` and explicitly scopes out documents with millions of entities. A large CSV is exactly one such document, so `update()` on it hits a documented scaling ceiling. Recorded as an open question.

## Contents

Proposals #1–#7 (the POC) and #8–#12 (follow-ups) in build order, a table mapping them to #82's acceptance criteria, a worked acceptance-scenario example with the traversal that answers it, a rejected-alternatives table with reasons, 8 open questions, and a P1–P5 phasing appendix.

## Verification

`mkdocs build --strict` produces no new warnings — the one remaining warning (`incremental-updates.md`) reproduces identically on a clean tree.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive structured data ingestion design covering tabular, nested, document-embedded, and existing-graph sources.
  * Documented deterministic mappings, identity resolution, relationships, provenance, updates, deletions, conflict handling, and retrieval compatibility.
  * Clarified that existing ingestion documentation applies to unstructured sources.
  * Added the structured ingestion proposal to the **Design Proposals** navigation.
  * Added spike findings and validation notes covering streaming, identity reconciliation, record chunks, and pipeline behavior.

### DIFF
--- a/docs/design/structured-ingestion.md
+++ b/docs/design/structured-ingestion.md
@@ -1,6 +1,9 @@
 # Design: Structured Data Ingestion
 
-**Status:** Proposed · **Tracking:** [FalkorDB/research#82][i82] (POC) · design from [research#65][i65] · supersedes [GraphRAG-SDK#74][i74]
+**Status:** Proposed, spike-validated · **Tracking:** [FalkorDB/research#82][i82] (POC) · design from [research#65][i65] · supersedes [GraphRAG-SDK#74][i74]
+
+The proposals in §3 were tested by five throwaway spikes against a live FalkorDB before this
+document settled — see [§10](#10-spike-results). Four of them corrected something here.
 
 [i82]: https://github.com/FalkorDB/research/issues/82
 [i65]: https://github.com/FalkorDB/research/issues/65
@@ -117,10 +120,27 @@ class RecordLoaderStrategy(ABC):
     async def load_records(self, source: str, ctx: Context) -> RecordBatch: ...
 
 class RecordBatch(DataModel):
-    records: Iterable[dict[str, Any]]   # streamed, never fully materialised
+    open_records: Callable[[], Iterator[dict[str, Any]]]  # a stream *factory* — see below
     document_info: DocumentInfo
     inferred_types: dict[str, str]      # column -> STRING/INTEGER/... hint from the reader
+    record_count: int | None = None     # when the loader knows it cheaply; None when streaming
+
+    def __iter__(self) -> Iterator[dict[str, Any]]:
+        return self.open_records()
 ```
+
+!!! warning "A factory, not an iterable — [spike s1][s1] corrected this"
+    The obvious signature `records: Iterable[dict[str, Any]]` **does not work**. Pydantic v2 keeps
+    it lazy (good: 200k rows cost ~0 MB vs 71.6 MB materialised) but replaces it with a *one-shot*
+    `ValidatorIterator`. #6 iterates records twice — step 3 builds record chunks, step 4 maps
+    records to `GraphData` — and the measured result is `step 3 saw 10 records, step 4 saw 0`, with
+    **no error raised**: a silent zero-row ingest. The annotation also erases list-ness, so `len()`
+    raises even when the caller passed a list, leaving no cheap count for progress reporting.
+    A factory is re-iterable by construction and verified `model_dump()`-safe.
+
+    Loaders therefore hand over a *re-openable* source (reopen the file, re-run the cursor). Where a
+    source genuinely cannot be read twice, the loader spools once and closes over the buffer — which
+    makes the memory cost explicit at the loader instead of silently corrupting the write.
 
 **Why.** Without it we get one bespoke code path per format — the exact problem #82 is filed
 against. With it, a new format is a ~50-line loader and *nothing else changes*. It also dissolves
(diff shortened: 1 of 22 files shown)


galshubeli:
## Spike round: five POCs, four corrections to the design

Following the suggestion to POC each proposal before building, I added `poc/structured-ingestion/` — five throwaway spikes, each answering **one** open question against the real `GraphStore`, the real `IngestionPipeline` and a live FalkorDB. No LLM, no API keys. `python run_all.py` → all 5 pass.

`poc/` is outside the wheel, outside `testpaths`, and outside CI (every job runs with `working-directory: graphrag_sdk` and lints only `src/`), so it ships nothing. **No `src` changes in this commit.**

**Four of the five falsified something in the design.** One inverted a headline decision.

| Spike | Outcome |
| --- | --- |
| s1 record stream | **Amended #1** — pydantic keeps `Iterable[dict]` lazy but *one-shot*. #6 iterates records twice (step 3 chunks, step 4 mapping); measured `step 3 saw 10 records, step 4 saw 0` — a **silent zero-row ingest**. Now a stream factory |
| s2 mapping DSL | **Amended #2** — label-addressed edges can't express a record with two nodes of the same label. `transactions.csv` (buyer + seller) produced a silent **self-loop**. Nodes get an `alias` |
| s3 identity | **Inverted #3** — see below |
| s4 record-as-chunk | **Confirmed #5**, including the predicted data-loss trap |
| s5 pipeline seam | **Amended #6** — share a base class, don't subclass; suppress `NEXT_CHUNK` for records |

### The one that matters most: `identity=["name"]` was wrong

| Policy | Acme nodes | #82 traversal |
| --- | --- | --- |
| name-first (**the design's default**) | 2 | **0 people reachable** |
| key-only | 2 | **0 people reachable** |
| key + `alias_ids` | **1** | **2 people reachable** |

`employees.csv` is an ordinary normalised table: it references its org by `org_id=ORG-42` and has **no `org_name` column**. Under name-first identity the mapping *cannot compute the identity of the entity it points at*, so #2's rule "every mapping supplies the type's identity attributes" is unsatisfiable for **any foreign key** — the most common structured shape there is. The failure is silent: a stub node collects all the `WORKS_AT` edges while the real Acme node holds the prose.

So structured writes are now **key-identified**, and `AliasMatchResolution` (#4) moves from optional nicety to **critical path**. Measured caveat: with no source carrying key *and* name together the bridge has nothing to build from — **ingest order doesn't matter, but presence does**, and unbridged stubs should be reported rather than left silent (new §8.9).

### The predicted trap is real

The review *predicted* from reading code that deterministic chunk uids keyed on the canonical document id would destroy data. Run against FalkorDB through the real `GraphStore`:

| chunk uid keyed on | before `update()` | shared with pending | **after cutover** |
| --- | --- | --- | --- |
| canonical doc id | 3 | 3 | **0** |
| effective (pending) uid | 3 | 0 | **3** |

The pending run `MERGE`s onto the live document's chunks, `rollforward_cutover()` deletes them, and an empty document is promoted — **no exception**. Today's `uuid4()` uids are accidentally immune, which is precisely why making them deterministic is the risky part of #5. Everything else in #5 held: `get_document_entity_candidates()`, `delete_stale_relationships()` and `delete_orphan_entities()` all behaved exactly as designed over record chunks.

### What didn't change

Record-as-chunk, `RELATES` + `rel_type`, mapping-as-ontology-fragment, deterministic no-LLM mapping, and "no retrieval strategy is touched" all survived contact with the database. Every correction is to a **signature or a default**, not to the shape of the design — which is the outcome a spike round should produce.

Details: [`FINDINGS.md`](https://github.com/FalkorDB/GraphRAG-SDK/blob/galshubeli-structured-ingestion-design/poc/structured-ingestion/FINDINGS.md) and each spike's `NOTES.md`. Design doc updated throughout, with a new §10 summarising the results.

galshubeli:
Closing — this was opened prematurely on my side. Keeping the structured-ingestion design and the spike round local for now; will re-open a PR when it's actually ready to review.

PR #297, opened 2026-08-13
Migrate to mintlify

### LINKED ISSUE #296: Migrate docs to mintlify

dudizimber:
fix #296 

- Created `storage.mdx` detailing GraphStore, VectorStore, and EntityDeduplicator functionalities.
- Added `strategies.mdx` outlining the Strategy pattern implementations for various concerns.
- Deleted `mkdocs.yml` configuration file.
- Introduced `check_mdx.mjs` script to validate MDX syntax across documentation files.

## Summary by CodeRabbit

- **Documentation**
  - Migrated documentation from MkDocs to Mintlify and published it on the FalkorDB documentation site.
  - Converted pages to MDX with improved metadata, navigation, branding, and SEO settings.
  - Updated README, project, and documentation links to use published site routes.
  - Added authoring guidance and local development instructions.

- **Quality Improvements**
  - Added automated MDX syntax and internal-link validation.
  - Documentation checks run for relevant pull requests, `main` updates, and manual requests.


galshubeli:
Thanks @dudizimber

scripts/check_mdx.mjs
```diff
@@ -0,0 +1,39 @@
...
+  return out;
+}
+
+const files = await walk(root);
+let failures = 0;
+
+for (const file of files) {
+  const source = await readFile(file, "utf8");
+  try {
+    await compile(source, { remarkPlugins: [remarkGfm] });
```
galshubeli: Missing `remark-frontmatter`, so the YAML block gets compiled as MDX content rather than stripped the way Mintlify strips it. That makes valid front matter fail CI.

Reproduced on a page whose only unusual content is `description: "Use <Note> for callouts"`:

```
docs/zz1.mdx:2:1 Expected a closing tag for `<Note>` (3:19-3:25) before the end of `setextHeading`
```

Given `docs/README.md` actively recommends `<Note>` / `<Tip>` / `<Warning>`, someone will hit this — and the reported position points at the front matter rather than the real location, which makes it a confusing failure to debug.

```js
import remarkFrontmatter from "remark-frontmatter";
...
await compile(source, { remarkPlugins: [remarkGfm, remarkFrontmatter] });
```

Plus `remark-frontmatter` on the workflow's `npm install --no-save` line.

Worth saying the script is otherwise doing its job — I confirmed it catches real errors (unclosed `<br>`, unbalanced `{`), and reporting every failure instead of stopping at the first is a genuine improvement over `mint`'s own checker.

docs/docs.json
```diff
@@ -0,0 +1,146 @@
...
+      "linkedin": "https://www.linkedin.com/company/falkordb",
+      "discord": "https://discord.com/invite/6M4QwDXn2w"
+    }
+  },
+  "integrations": {
+    "gtm": {
+      "tagId": "GTM-MBWB627H"
+    }
+  },
+  "redirects": [
```
galshubeli: These 16 redirects can't ever fire, for two independent reasons.

MkDocs used directory URLs (`use_directory_urls` defaults to true and was never overridden), so `.html` is a URL form the old site never served. Against the still-live Pages deployment:

```
/GraphRAG-SDK/getting-started       -> 301
/GraphRAG-SDK/getting-started/      -> 200
/GraphRAG-SDK/getting-started.html  -> 404
```

And separately, those URLs live on `falkordb.github.io` — a different origin, where a redirect configured on `docs.graphrag.falkordb.com` has no reach regardless of path shape.

Suggest deleting the block. If preserving old inbound links genuinely matters, it has to happen on the Pages side instead (meta-refresh stubs per page), which is also the only thing that would make unpublishing Pages non-lossy.

.github/workflows/docs.yml
```diff
@@ -1,47 +1,40 @@
-name: Deploy Docs
+name: Docs
+
+# Mintlify deploys docs/ automatically from the default branch via its GitHub
```
galshubeli: Two things worth confirming against this comment, since both are invisible from the diff:

1. The deployment branch is currently set to this feature branch, not the default branch — `docs.graphrag.falkordb.com` is already serving content that only exists in this PR (the 4-product switcher, the "Try GraphRAG" button). Needs repointing to `main` after merge.
2. Removing the deploy job doesn't unpublish the old site. `has_pages` is still `true` with `build_type: workflow`, so `falkordb.github.io/GraphRAG-SDK/` keeps serving its last build forever. Worth disabling Pages in repo settings as part of this migration — otherwise there are two live doc sites, and `pyproject.toml:77` still points at the stale one.

.github/workflows/docs.yml
```diff
@@ -1,47 +1,40 @@
...
-  deploy:
-    needs: build
-    runs-on: ubuntu-latest
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
-    steps:
-      - id: deployment
-        uses: actions/deploy-pages@v5
+          node-version: "lts/*"
```
galshubeli: `node-version: "lts/*"` here and the unpinned `npm install -g mint` on line 39 both mean an upstream release can break this job with no change on your side — and since it gates docs correctness rather than the package build, that failure will land at an unrelated moment. Worth pinning both to explicit versions.

docs/README.md
```diff
@@ -0,0 +1,72 @@
...
+mint dev            # preview at http://localhost:3000
+mint broken-links   # verify every internal link resolves
+```
+
+`mint` requires an LTS release of Node.
+
+To validate MDX syntax across every page in one pass:
+
+```bash
+npm install --no-save @mdx-js/mdx remark-gfm
```
galshubeli: Following this locally leaves an untracked `node_modules/` (and `package-lock.json`) at the repo root — `.gitignore` doesn't cover either. Worth adding `node_modules/` to `.gitignore` alongside this instruction.

galshubeli (commented):
Reviewed this by checking the branch out locally and validating against the live deployment. The migration itself is careful work — flagging what I found, plus a few things that live outside the diff.

**Verified clean:** no MkDocs-only syntax left to port (zero `!!!` admonitions, `=== "Tab"` blocks, or mermaid fences, despite `mkdocs.yml` having configured all three); no H1 in any body and no heading-level jumps; all 16 pages registered in `navigation` with zero orphans and zero broken internal links; `docs.json` validates against `https://mintlify.com/docs.json` with 0 errors; content fidelity is exact (3–9 changed lines per page, all front matter or link rewrites). I also confirmed the underscore-bearing anchors (`#delete_document`, `#apply_changes`, `#concurrency-and-the-update_concurrency-trap`) resolve against the deployed HTML — MkDocs and Mintlify slugify those identically.

Inline comments below. Not duplicating what CodeRabbit and Copilot already caught.

---

### Two blocking items I can't anchor inline (files aren't in the diff)

**1. `graphrag_sdk/README.md:167,181-187` — 8 links will 404 on PyPI after merge.**

`graphrag_sdk/pyproject.toml:9` sets `readme = "README.md"`, so this file is the published PyPI long description. It links to `blob/main/docs/benchmark.md`, `getting-started.md`, `architecture.md`, `configuration.md`, `strategies.md`, `providers.md`, and `api-reference.md`. Those all return 200 today and break the moment the `.md` → `.mdx` renames land on `main`. The root `README.md` was updated to the new domain; this one was missed.

**2. `graphrag_sdk/pyproject.toml:77` — `Documentation = "https://falkordb.github.io/GraphRAG-SDK/"`.**

Should be `https://docs.graphrag.falkordb.com`. See the Pages note below for why this one compounds.

---

### Operational, outside the diff

**3. Production docs are already live off this unmerged branch.** `docs.graphrag.falkordb.com` currently serves the 4-product switcher, the "Try GraphRAG" button, and the `Strategy Reference` sidebar title — all of which exist only in this PR's `docs.json`. So Mintlify's deployment branch is pointed at `296-migrate-docs-to-mintlify`, not `main`. Worth repointing to `main` after merge and confirming a fresh build, otherwise the site quietly freezes at this branch's state.

**4. GitHub Pages is still published.** The API reports `has_pages: true`, `build_type: workflow`, source `main`. Deleting the deploy workflow stops updates but doesn't unpublish — `https://falkordb.github.io/GraphRAG-SDK/getting-started/` returns 200 right now and will serve a frozen copy indefinitely. Combined with `"seo": {"indexing": "all"}` on the new site, that's two indexed copies of the same docs competing in search. Worth turning Pages off in repo settings as part of this change.

**5. `validate` isn't a required status check.** Since the Mintlify GitHub app publishes independently of Actions, the MDX/link validation is currently advisory — a bad merge is live before CI finishes.

---

### One design question: is the `.md` → `.mdx` rename worth it?

Mintlify renders `.md` natively (its own docs recommend it to "expedite migration from other platforms"), and across all 16 pages the count of Mintlify components actually used is zero — no `<Note>`, `<CodeGroup>`, `<Accordion>`, nothing. So the rename buys no capability today, and keeping `.md` preserves the option of renaming individual pages later when one actually needs JSX.

What it costs is permanent: every inbound link to `github.com/FalkorDB/GraphRAG-SDK/blob/main/docs/*.md` — from issues, blog posts, other repos — breaks with no redirect available, since GitHub blob URLs have no redirect mechanism and `docs.json` redirects don't reach them. Item 1 above is that cost arriving on day one.

Entirely your call, and not something I'd hold the PR for — but this is the cheapest moment to make it.

scripts/check_mdx.mjs
galshubeli: Confirmed fixed, and I checked the behaviour rather than just the diff: my original repro (`description: "Use <Note> for callouts"`) now compiles clean, and the checker still catches real errors — `line<br>` still fails with `Expected a closing tag for <br>`. Plugin order is right too.

Pinning the three packages to exact versions on the install line was a better call than what I suggested.

galshubeli:
Re-reviewed at `30af631`. All five inline items are fixed, and the version pinning went further than I asked (`node 24.19.0`, `@mdx-js/mdx@3.1.1`, `remark-frontmatter@5.0.0`, `remark-gfm@4.0.1`, `mint@4.2.804`). MDX check is 16 files / 0 errors, navigation still has no orphans or dangling entries.

I also read FalkorDB/docs#544 this time, which I should have done before commenting. Two things in my review were wrong as a result.

**Withdrawing the `.mdx` objection.**

> ### One design question: is the `.md` → `.mdx` rename worth it?
>
> Mintlify renders `.md` natively (its own docs recommend it to "expedite migration from other platforms"), and across all 16 pages the count of Mintlify components actually used is zero [...] So the rename buys no capability today

This was reasoning about a standalone site, which isn't what's being built. `scripts/aggregate_docs.py` mounts each source repo's `docs/` into one Mintlify deployment, so `.mdx` is what the aggregator consumes — the rename is load-bearing, not speculative. Ignore that section.

I'd also flag that the design answers the concern I would have raised about the `/graphrag` mount: `rewrite_internal_links(root, mount, known)` prefixes only the paths a repo owns, so `](/configuration)` becomes `/graphrag/configuration` while cross-product links are deliberately left alone, and the aggregated build re-runs `mint validate` + `mint broken-links` over the merged site. Relatedly, the full `navigation.products` block in this repo's `docs.json` is the contract `own_product()` requires, not leftover standalone config — I'd misread it as the latter.

**Superseding my operational item 3.**

> **3. Production docs are already live off this unmerged branch.** [...] So Mintlify's deployment branch is pointed at `296-migrate-docs-to-mintlify`, not `main`. Worth repointing to `main` after merge

Obsolete under the aggregation model — Mintlify deploys FalkorDB/docs' `docs-site` branch, and `sources.json` pulls this repo's `docs/` from `main`, so there's no per-repo deployment branch to repoint. Disregard.

---

Three things I think are still worth doing.

**1. `graphrag_sdk/README.md:167,181-187` is still unfixed** — this was in my review body rather than an inline comment, so there was no thread to mark resolved, which is probably why it slipped:

> `graphrag_sdk/pyproject.toml:9` sets `readme = "README.md"`, so this file is the published PyPI long description. It links to `blob/main/docs/benchmark.md`, `getting-started.md`, `architecture.md`, `configuration.md`, `strategies.md`, `providers.md`, and `api-reference.md`. Those all return 200 today and break the moment the `.md` → `.mdx` renames land on `main`.

Re-confirmed on `30af631` — all 8 still point at `blob/main/docs/*.md`.

**2. Merge order matters more than it looks.** `docs.falkordb.com/graphrag` returns 404 today, and `repository_dispatch` only triggers workflows on a repository's *default* branch — so until #544 merges, `notify-docs-repo` will dispatch successfully and nothing will consume it. Meanwhile `README.md`, `pyproject.toml`, `CHANGELOG.md` and `docs/README.md` all point at that 404. The `pyproject.toml` one is the sharp edge: any release cut in the gap ships a dead `Documentation` URL in its PyPI metadata. Landing #544 first (or both together) avoids the whole window.

**3. GitHub Pages is still published**, which was operational item 4 and needs a settings change rather than a code one:

> The API reports `has_pages: true`, `build_type: workflow`, source `main`. Deleting the deploy workflow stops updates but doesn't unpublish — `https://falkordb.github.io/GraphRAG-SDK/getting-started/` returns 200 right now and will serve a frozen copy indefinitely.

Still `true` as of today. And one adjacent decision worth making deliberately: `docs.graphrag.falkordb.com` is live serving these docs and is registered as this product's `site` in `sources.json`, so once the aggregate is up there are two indexed copies of identical content under `"seo": {"indexing": "all"}`. Redirecting it to `/graphrag` or declaring it canonical are both fine — just not neither.

galshubeli (approved):

PR #305, opened 2026-08-23
fix(retrieval): one LLM repair round for rejected text-to-Cypher queries

### LINKED ISSUE #292: Text-to-Cypher: invalid generated queries are retried 3x and never repaired
## Summary

`validate_cypher()` ([`retrieval/strategies/cypher_generation.py:347`](https://github.com/FalkorDB/GraphRAG-SDK/blob/main/graphrag_sdk/retrieval/strategies/cypher_generation.py#L347)) is a **safety allowlist, not a syntax check**. It verifies the query starts with a read-only keyword, contains a `RETURN`, has no writes / `CALL` / `LOAD CSV`, and references only known labels. Nothing validates Cypher grammar.

Syntactically invalid generated queries therefore pass validation and fail inside FalkorDB at execution time:

```
Invalid input 'F': expected OR or OPTIONAL MATCH
The alias 'r' was specified for both a node and a relationship
Unexpected clause following RETURN
```

## Observed

GraphRAG-Bench Medical, 2062 questions, `gpt-4o-mini`, `enable_cypher=True`, SDK 1.3.0:

| Run | Failed Cypher queries | Rate |
| --- | ---: | ---: |
| 1 | 59 | 2.9% |
| 2 | 30 | 1.5% |

Each failure is logged as `FalkorDB query failed after 3 attempts: ResponseError`.

## Two problems

**1. Deterministic errors are retried 3x.** A syntactically malformed query cannot succeed on retry — the input is identical each time. That is 90–177 wasted round trips per run, on the critical path of a user's question.

**2. No repair attempt.** The FalkorDB error message is specific and actionable (`Unexpected clause following RETURN`). Feeding it back to the LLM for a single repair pass would recover most of these. Today the Cypher path silently contributes nothing for those questions.

## Impact

Latency and lost recall, not correctness. `MultiPathRetrieval`'s other paths cover the gap — both benchmark runs produced complete answers with 0 errors and 0 empty answers — which is exactly why this is invisible in normal use.

Note this is on the experimental path (`enable_cypher` defaults to `False`), but it is enabled in the configuration behind our published [GraphRAG-Bench results](https://github.com/FalkorDB/GraphRAG-SDK/blob/main/docs/benchmark.md), so it is running in a configuration we support and advertise.

## Suggested fixes

1. Do not retry on `ResponseError` where the message indicates a syntax/semantic error — fail fast instead. Cheap, removes ~2/3 of the waste immediately.
2. On a syntax failure, send the query plus the FalkorDB error back to the LLM for one repair attempt before giving up.
3. Optionally, count and surface Cypher-path failures in retrieval metadata so the rate is observable rather than log-only.

lingmao233:
Fixes #292 (repair round + remaining fail-fast classification).

## What

Text-to-Cypher queries that pass the safety allowlist but get rejected by FalkorDB currently degrade silently: `execute_cypher_retrieval` catches the execution error and returns empty results, and the two error classes from the benchmark runs in #292 that are not in `_NON_TRANSIENT_MARKERS` still burn the full 3-attempt retry budget on identical input.

1. **One repair round** — when execution fails, `execute_cypher_retrieval` now sends the failed query plus the FalkorDB error back to the LLM for exactly one corrected attempt (`_repair_cypher`). The repaired query must pass the same `validate_cypher` read-only allowlist as a fresh generation, so the repair path cannot smuggle in writes. If the repaired query also fails, the path degrades to empty results exactly as before — no retry loop.
2. **Fail-fast for the remaining deterministic errors** — `_NON_TRANSIENT_MARKERS` now also covers the two error classes observed in #292 that were not classified yet: alias-reused-for-node-and-relationship and unexpected-clause.

The optional observability suggestion from #292 (surfacing per-path failure stats) is not included here.

## Why

Per #292's benchmark runs, 1.5–2.9% of generated queries failed deterministically; every failure cost a full 3x retry on identical bytes (90–177 wasted round trips per run) plus a 100% recall loss on those questions, because no repair was ever attempted. FalkorDB's errors are specific enough that one LLM repair round should recover most of them.

## How I verified

New tests in `graphrag_sdk/tests/test_cypher_generation.py` (mock LLM + mock graph store, no live services):

- repair round is attempted exactly once with the server error included in the prompt, and a successful repair produces results
- a repair that fails again returns empty results with no retry loop (exactly 2 executions)
- a "repaired" write query is blocked by `validate_cypher` and never reaches the database
- the two newly classified error strings (plus the previously covered ones) are non-transient, while connection errors remain transient

The existing suite is unchanged — the silent-degradation test now exercises the repair path and still returns empty results.

No change to the success path: generation and first execution behave identically, and the repair only activates after a failure, so benchmark accuracy is unaffected by construction. Happy to run the 100-question benchmark if maintainers want the numbers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of invalid or unsupported Cypher queries.
  - Automatically attempts one safe query correction when execution fails.
  - Prevents unsafe corrected queries from being executed.
  - Returns an empty result instead of failing when a query cannot be safely repaired.
  - More accurately distinguishes permanent query errors from temporary connection issues, reducing unnecessary retries.

### DIFF
--- a/graphrag_sdk/src/graphrag_sdk/core/connection.py
+++ b/graphrag_sdk/src/graphrag_sdk/core/connection.py
@@ -249,6 +249,10 @@ class FalkorDBConnection:
         "unknown function",
         "type mismatch",
         "procedure not found",
+        # Semantic rejections observed on LLM-generated queries (#292) —
+        # deterministic, so a retry can only replay the same failure.
+        "for both a node and a relationship",
+        "unexpected clause",
     )
 
     @classmethod
--- a/graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_generation.py
+++ b/graphrag_sdk/src/graphrag_sdk/retrieval/strategies/cypher_generation.py
@@ -478,6 +478,55 @@ async def generate_cypher(
     return None
 
 
+async def _repair_cypher(
+    llm: Any,
+    cypher: str,
+    error: str,
+    *,
+    ontology: Ontology | None = None,
+    ctx: Context | None = None,
+) -> str | None:
+    """Ask the LLM for one corrected query after a FalkorDB execution error.
+
+    The repaired query must pass the same ``validate_cypher`` safety
+    allowlist as a fresh generation — a repair round must never become a
+    bypass around the read-only checks. Returns ``None`` when the LLM
+    call fails, the answer is empty/unparseable, or the proposal fails
+    validation.
+    """
+    prompt = (
+        "The following read-only Cypher query failed when executed on FalkorDB.\n"
+        "Fix it. Keep it read-only (no CREATE/DELETE/SET/MERGE/REMOVE/CALL),\n"
+        "give every RETURN column a unique alias, and keep a LIMIT clause.\n\n"
+        f"Failed query:\n```cypher\n{cypher}\n```\n\n"
+        f"FalkorDB error:\n{error}\n\n"
+        "Return ONLY the corrected Cypher query inside triple backticks."
+    )
+    try:
+        if ctx is not None:
+            ctx.ensure_budget("Cypher repair LLM call")
+        response = await llm.ainvoke(
+            prompt,
+            timeout=(
+                ctx.provider_timeout_seconds("Cypher repair LLM call")
+                if ctx is not None
+                else None
+            ),
+        )
+    except LatencyBudgetExceededError:
+        raise
+    except Exception as exc:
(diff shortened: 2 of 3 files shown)


galshubeli:
@lingmao233 thanks for your contribution! please fix the ci

hd-o:
@lingmao233 are you available? or @galshubeli can I open a new PR with the fixes included?

Issue #3, opened 2024-06-25
Support multi agent architecture


dudizimber:
Introduce Orchestrator agent
Introduce the ability to **register** a **specialist** agent

PR #309, opened 2026-08-25
Chunking: idempotent re-ingest and a measured 384-token default

Naseem77:
## Chunking — three fixes

Measured on an 11-document benchmark against a real FalkorDB. Part of FalkorDB/research#88.

### 1. Re-uploading a file duplicated its chunks
**Problem:** every ingest gave chunks fresh random ids, so `MERGE` never matched — 17 → 34 → 51 chunks over three uploads of the same file.
**Fix:** chunk id is derived from document id + position + text (`_assign_deterministic_chunk_uids`). Same chunk → same id → stored once. 17 → 17 → 17. `ContextualChunking` hashes the original chunk text, not the LLM-enriched one.

### 2. Default chunk size 512 → 384 tokens
**Problem:** 512 was an unmeasured default; quality falls as chunks grow (relation F1 0.233 / 0.221 / 0.204 at 384 / 768 / 1536).
**Fix:** 384 in `SentenceTokenCapChunking`, `StructuralChunking`, `ContextualChunking`, the `CallableChunking` example, the `GraphRAG` docstring and the docs. Entity F1 0.574 vs 0.563, relation F1 0.237 vs 0.223 against 768. **Cost:** ~35 % more extraction LLM calls, ~19 % more input tokens per ingest (stated in the CHANGELOG). Re-measured on top of #310's prompt with the full stack, 3 repeats, against main with its LLM resolver on: exact relation F1 0.062 → 0.134, answer accuracy 27.0 → 32.0 %, QA input tokens −18 %, for +$0.03 per run.

### 3. Re-uploading an unchanged file still ran the whole pipeline
**Problem:** fix 1 kept the chunk count flat, but chunking, NER, the LLM calls and the graph writes still ran — 33 provider calls and ~30 s per re-upload, relations grew (105 → 121 → 129) and existing edge types were rewritten.
**Fix:** right after load, hash the document text and compare with the stored `Document` node's `content_hash`. Same hash → return before chunking with `IngestionResult.metadata.skipped_unchanged=True`, `chunks_indexed=0`. Measured: 0 calls, 0.01 s, 0 new chunks / entities / relations. The hash is written only after a run completes, so a partially failed ingest is retried in full. A changed file still takes the full path.

### Review follow-up (d38cf36)
- Fixes 1 and 3 now work for direct `IngestionPipeline.run()` too, not only via `GraphRAG.ingest()`: when no `document_info` is given the pipeline derives the Document id from the normalised source path (text mode: a text hash), mirroring `GraphRAG._resolve_document_id`. `update()` passes an explicit id and is unaffected.
- Unchanged-document check moved before chunking; `chunks_indexed` reports 0.
- `content_hash` written after steps 8/9, not in step 3.
- Dead `uid=sc.uid` in `structural_chunking.py` removed; docs updated to 384; CHANGELOG covers both fixes, the 384 cost, the one-time extra chunk copy on first re-ingest after upgrade, and edited-file re-ids.
- New tests: two `run()`s of the same file → same chunk ids (no `document_info`); same text in two files does not collide; text mode stable; chunker not called on skip; skip fires with a path-derived id; no hash after a failed extraction; contextual `original_chunk` hashing.

### Verification
Full suite on this branch: **1112 passed, 41 skipped**. Live re-ingest probe on FalkorDB: 0 provider calls, chunks / entities / RELATES unchanged, `content_hash` present on every Document after a successful run.

Refs: FalkorDB/research#88, FalkorDB/research#71

## Summary by CodeRabbit

* **New Features**
  * Unchanged documents now skip unnecessary reprocessing during re-ingestion.
  * Document and chunk identifiers remain stable across repeated runs, including deterministic IDs for text ingestion.
  * Filesystem paths are normalized while URI sources remain unchanged.

* **Improvements**
  * Default chunk size is now 384 tokens; explicit overrides remain available.
  * Contextual chunk updates preserve identifiers for unchanged source content.
  * Failed ingestion runs do not record a content hash.

* **Documentation**
  * Updated guides, examples, benchmarks, and changelog with the revised defaults and behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Re-ingesting the same file no longer duplicates its chunks. Chunk ids are
+  now derived from document id + position + text instead of a fresh
+  `uuid4()`, so `MERGE` finds the existing node (17 → 17 → 17 chunks across
+  three ingests, previously 17 → 34 → 51). `ContextualChunking` hashes the
+  original chunk text, not the LLM-enriched one. Applies to
+  `IngestionPipeline.run()` directly as well as through `GraphRAG.ingest()`:
+  when no `document_info` is supplied the pipeline now derives a stable
+  Document id from the normalised source path (or, in text mode, from the
+  text). **Upgrade note:** graphs built before this change hold random chunk
+  ids; the first re-ingest of an existing document adds one more copy of its
+  chunk layer (matching nothing), and is stable from the second re-ingest on.
+  Because position is part of the id, inserting a paragraph into an edited
+  document re-ids every later chunk; only byte-identical files are a no-op.
+- Re-ingesting an unchanged document is now a true no-op. The pipeline hashes
+  the loaded text and, if the stored Document carries the same
+  `content_hash`, returns before chunking with
+  `IngestionResult.metadata["skipped_unchanged"] = True` and
+  `chunks_indexed = 0` — no chunker, NER, LLM or graph calls (measured: 33
+  provider calls and ~30 s → 0 and 0.01 s). The hash is written only after a
+  run completes, so a partially failed ingest is retried in full rather than
+  skipped.
+
+### Changed
+
+- Default chunk size lowered from 512 to 384 tokens in
+  `SentenceTokenCapChunking`, `StructuralChunking`, `ContextualChunking` and
+  the documented `CallableChunking` example. Measured on the benchmark corpus:
+  entity F1 0.574 vs 0.563 and relation F1 0.237 vs 0.223 against 768; with
+  the current extraction prompt, exact relation F1 2.2× and answer accuracy
+  27 → 32 % for the full stack. **Cost:** ~35 % more extraction LLM calls and
+  ~19 % more input tokens per ingest (103 → 157 calls on the 11-document
+  corpus), since the per-call instructions are re-sent once per chunk. Pass
+  `max_tokens=512` to keep the old size.
 - Fixed vector-search ordering so chunk, entity, and relationship searches use
   similarity scores, with higher values indicating closer matches.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,19 +41,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   same text hash (previously a fresh `text-<uuid>` per call), so ingesting
   identical text twice is now one document and a no-op the second time; pass
   `document_id` to store identical text as distinct documents.
+  **Breaking:** code that called `ingest(text=...)` once per record and
+  relied on every call creating its own Document now gets one Document per
+  distinct text and a zero-result `skipped_unchanged` no-op for each
+  duplicate — pass a `document_id` per record to keep the old behaviour.
+  A caller-supplied `document_info` is merged field by field using
+  `model_fields_set`, so `DocumentInfo(path=...)` without an explicit `uid`
+  takes the derived id instead of the model's random default. Ids the
+  pipeline derives from a source path are checked for the reserved
+  `__pending__` marker (`ValueError`), the same rule `GraphRAG` applies to
+  explicit ids.
   **Upgrade note:** graphs built before this change hold random chunk
   ids; the first re-ingest of an existing document adds one more copy of its
   chunk layer (matching nothing), and is stable from the second re-ingest on.
   Because position is part of the id, inserting a paragraph into an edited
   document re-ids every later chunk; only byte-identical files are a no-op.
+  Chunks written by `update()` are keyed on its transient pending id (the
+  pending and live documents must not share chunk nodes during the
+  cutover); the `content_hash` the cutover records is what keeps a later
+  `ingest()` of that content a no-op.
 - Re-ingesting an unchanged document is now a true no-op. The pipeline hashes
   the loaded text and, if the stored Document carries the same
   `content_hash`, returns before chunking with
   `IngestionResult.metadata["skipped_unchanged"] = True` and
   `chunks_indexed = 0` — no chunker, NER, LLM or graph calls (measured: 33
   provider calls and ~30 s → 0 and 0.01 s). The hash is written only after a
-  run completes, so a partially failed ingest is retried in full rather than
-  skipped.
+  run completes with every write reported in full: a failed extraction, a
+  `RELATES`/`MENTIONED_IN` edge the graph store dropped after a transient
+  error, or a chunk left without an embedding by a rate-limited embedder all
+  withhold the hash (`IngestionResult.metadata["incomplete_writes"]` lists
+  the shortfalls), so the next ingest repairs the document instead of
+  skipping it. Graph-store adapters without `get_document_record` keep
+  working — the pipeline skips the check instead of raising.
+- `GraphRAG.update(..., force=True)` re-extracts a document whose content
+  hash is unchanged. With `ingest()` now skipping unchanged documents, this
+  is the supported way to re-chunk or re-extract existing text after
+  changing the ontology, chunker, extractor or model (for example to adopt
+  the new 384-token default below); `update_sync()` takes the same flag.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,8 +71,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   error, or a chunk left without an embedding by a rate-limited embedder all
   withhold the hash (`IngestionResult.metadata["incomplete_writes"]` lists
   the shortfalls), so the next ingest repairs the document instead of
-  skipping it. Graph-store adapters without `get_document_record` keep
-  working — the pipeline skips the check instead of raising.
+  skipping it. `update()` applies the same gate: its cutover promotes the
+  pending Document without a hash when the pipeline reported a shortfall
+  (and crash recovery of such a pending rolls forward uncertified rather
+  than refusing). Deployments without an embedder are not penalised —
+  `VectorStore.index_chunks` now returns `None` (nothing attempted) instead
+  of `0` (every embedding failed) when no embedder is configured, so
+  graph-only ingests still record the hash. Graph-store adapters without
+  `get_document_record` keep working — the pipeline skips the check instead
+  of raising.
 - `GraphRAG.update(..., force=True)` re-extracts a document whose content
   hash is unchanged. With `ingest()` now skipping unchanged documents, this
   is the supported way to re-chunk or re-extract existing text after
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   original chunk text, not the LLM-enriched one. Applies to
   `IngestionPipeline.run()` directly as well as through `GraphRAG.ingest()`:
   when no `document_info` is supplied the pipeline now derives a stable
-  Document id from the normalised source path (or, in text mode, from the
-  text). **Upgrade note:** graphs built before this change hold random chunk
+  Document id from the source (normalised filesystem path; URIs kept
+  verbatim), and in text mode from a hash
+  of the text. `GraphRAG.ingest(text=...)` without `document_id` uses the
+  same text hash (previously a fresh `text-<uuid>` per call), so ingesting
+  identical text twice is now one document and a no-op the second time; pass
+  `document_id` to store identical text as distinct documents.
+  **Upgrade note:** graphs built before this change hold random chunk
   ids; the first re-ingest of an existing document adds one more copy of its
   chunk layer (matching nothing), and is stable from the second re-ingest on.
   Because position is part of the id, inserting a paragraph into an edited
@@ -38,9 +43,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the documented `CallableChunking` example. Measured on the benchmark corpus:
   entity F1 0.574 vs 0.563 and relation F1 0.237 vs 0.223 against 768; with
   the current extraction prompt, exact relation F1 2.2× and answer accuracy
-  27 → 32 % for the full stack. **Cost:** ~35 % more extraction LLM calls and
-  ~19 % more input tokens per ingest (103 → 157 calls on the 11-document
-  corpus), since the per-call instructions are re-sent once per chunk. Pass
+  27 → 32 % for the full stack. **Cost:** more extraction LLM calls per
+  ingest — 103 → 157 (+52 %) on the 11-document benchmark corpus, +35 % on a
+  53k-token corpus — and ~19 % more input tokens, since the per-call
+  instructions are re-sent once per chunk. Pass
   `max_tokens=512` to keep the old size.
 - Fixed vector-search ordering so chunk, entity, and relationship searches use
   similarity scores, with higher values indicating closer matches.
--- a/docs/benchmark.mdx
+++ b/docs/benchmark.mdx
@@ -74,7 +74,7 @@ model and the judge.
 | Generation temperature | 0.7 | Appendix H.2 |
 | Framework | GraphRAG-SDK 1.3.0 (PyPI) on FalkorDB | — |
 | Graph layout | one graph per corpus document | — |
-| Chunking | `SentenceTokenCapChunking`, max_tokens 512, overlap 2 sentences | SDK default |
+| Chunking | `SentenceTokenCapChunking`, max_tokens 384 (was 512 before 1.2), overlap 2 sentences | SDK default |
 | Retrieval | `MultiPathRetrieval` — chunk_top_k 15, rel_top_k 15, max_entities 30, max_relationships 20, keyword_limit 10 | SDK default |
 | Embeddings | `text-embedding-3-large` @ 1024 dimensions | Declared below |
 | Text-to-Cypher | enabled | Declared below |
--- a/docs/benchmark.mdx
+++ b/docs/benchmark.mdx
@@ -74,7 +74,7 @@ model and the judge.
 | Generation temperature | 0.7 | Appendix H.2 |
 | Framework | GraphRAG-SDK 1.3.0 (PyPI) on FalkorDB | — |
 | Graph layout | one graph per corpus document | — |
-| Chunking | `SentenceTokenCapChunking`, max_tokens 384 (was 512 before 1.2), overlap 2 sentences | SDK default |
+| Chunking | `SentenceTokenCapChunking`, max_tokens 384 (was 512 before this release), overlap 2 sentences | SDK default |
 | Retrieval | `MultiPathRetrieval` — chunk_top_k 15, rel_top_k 15, max_entities 30, max_relationships 20, keyword_limit 10 | SDK default |
 | Embeddings | `text-embedding-3-large` @ 1024 dimensions | Declared below |
 | Text-to-Cypher | enabled | Declared below |
--- a/graphrag_sdk/.venv
+++ b/graphrag_sdk/.venv
@@ -1,0 +1,1 @@
+/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
--- a/graphrag_sdk/.venv
+++ b/graphrag_sdk/.venv
@@ -1,1 +1,0 @@
-/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -1369,10 +1369,13 @@ class GraphRAG:
 
         Uses sensible defaults for any unspecified strategy:
         - Loader: auto-detected from file extension (PDF or text)
-        - Chunker: SentenceTokenCapChunking(max_tokens=512, overlap_sentences=2)
+        - Chunker: SentenceTokenCapChunking(max_tokens=384, overlap_sentences=2)
           — sentence-aware, never splits entity names at chunk boundaries.
-          Override with ``chunker=FixedSizeChunking(...)`` if you need
-          character-window chunking.
+          384 is the measured best size for relationship extraction, which
+          returns a roughly fixed number of facts per call — smaller chunks
+          mean more calls and more facts. Override with
+          ``chunker=FixedSizeChunking(...)``
+          if you need character-window chunking.
         - Extractor: GraphExtraction with configured LLM
         - Resolver: ExactMatchResolution
 
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -38,6 +38,7 @@ from graphrag_sdk.core.models import (
     RagResult,
     RetrieverResult,
     UpdateResult,
+    stable_document_id,
 )
 from graphrag_sdk.core.providers import Embedder, LLMInterface
 from graphrag_sdk.discovery import SchemaExtensionProposal, suggest_extensions
@@ -1419,7 +1420,7 @@ class GraphRAG:
         if isinstance(source, list) and document_id is not None:
             raise ValueError(
                 "'document_id' cannot be set on batch ingest (list source). "
-                "Each file's id defaults to os.path.normpath(path); pass an "
+                "Each file's id defaults to its normalised path; pass an "
                 "explicit document_id only for single-source calls."
             )
         if text is not None and loader is not None:
@@ -1496,21 +1497,25 @@ class GraphRAG:
         """Compute the stable Document node id for an ingest/update call.
 
         - explicit ``document_id`` → used verbatim
-        - file mode (source given, no id) → ``os.path.normpath(source)``
-        - text mode (no id) → generated ``text-<8hex>``
+        - file mode (source given, no id) → ``stable_document_id(source)``:
+          the normalised path, or the URI verbatim
+        - text mode (no id) → ``text-<16hex>`` derived from the text
 
         Path normalization collapses ``./``, ``../``, and double slashes
         so the same logical path always yields the same id, regardless of
-        how the caller spelled it.
+        how the caller spelled it; sources with a URI scheme are not
+        touched, since ``normpath`` would rewrite them. Text mode hashes the text (SHA-256, 64
+        bits) so ingesting the same text twice is the same document — and
+        therefore a no-op on the second call, like a file — instead of a
+        fresh random ``text-<uuid>`` per call. Two different texts collide
+        with probability ~2 in 10^11 at 10K ingests; pass ``document_id``
+        when you need to ingest identical text as distinct documents.
         """
         if document_id is not None:
             return document_id
         if text is None and source is not None:
-            return os.path.normpath(source)
-        # 64-bit suffix — at 32 bits (the original [:8]), 10K text-mode
-        # ingests in one session collide with ~12% probability. 64 bits
-        # pushes that to roughly 2 in 10^11 for the same volume.
-        return f"text-{uuid4().hex[:16]}"
+            return stable_document_id(source)
+        return f"text-{hashlib.sha256((text or '').encode('utf-8')).hexdigest()[:16]}"
 
     async def _ingest_single(
         self,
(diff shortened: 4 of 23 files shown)


galshubeli:
## Review

Verified first: CI is green, I reproduced **1102 passed / 41 skipped** locally, and the branch merges cleanly onto current `main`.

The PR bundles two unrelated changes — stable chunk IDs, and the 512 → 384 default. I'd split them: the first is a bug fix that's nearly ready, the second is a tuning decision I think should wait.

---

### Chunk ID fix

- ✅ **The bug is real and the design is right.** Random `uuid4` identity for a node that gets `MERGE`d was a defect; deriving it from document + position + text is the correct answer.
- ❌ **It only takes effect through `GraphRAG.ingest()`.** `DocumentInfo.uid` defaults to `uuid4()` (`core/models.py:100`) and no loader sets it, so `doc_key = doc_info.uid or doc_info.path or ""` always takes the random uid and the `path` fallback is unreachable. Using `IngestionPipeline` directly still mints fresh IDs every run — verified, two runs produce two different ID sets. Only the facade works, via `main.py:1583`.
- ❌ **No test covers the real path.** All six new tests call `_assign_deterministic_chunk_uids` directly with a literal `doc_uid="doc-1"`, so none exercises the default. There's no re-ingest test at pipeline or facade level anywhere in the suite — the one behaviour this PR exists to fix is the one thing not covered end to end. One test that runs `pipeline.run()` twice and asserts equal uid sets would catch it.
- ⚠️ **Upgrade side effect worth documenting.** Existing graphs hold random chunk IDs. The first re-ingest after upgrade mints deterministic IDs that match nothing, so it adds one more full copy of the chunk layer — and those orphans keep their `PART_OF` edges, so they stay in retrieval as duplicate context. Stable only from the second re-ingest on.
- ⚠️ **Scope limit worth stating in the description.** Index and text are both in the hash, so inserting a paragraph near the top re-mints every downstream ID. "The count stays flat" holds for byte-identical files; edited files still grow the lexical layer under plain `ingest()`.
- 🧹 `structural_chunking.py:182` passes `uid=sc.uid` to preserve the fallback sub-chunk's identity — now dead, since the helper overwrites unconditionally.

**One note on the fix:** the obvious repair — keying on `doc_info.path` instead of `uid` — would break `update()`. `delete_document_chunks` does `MATCH (:Document {id})-[:PART_OF]->(c:Chunk) DETACH DELETE c` (`graph_store.py:657`); the pending and live documents share a path but not an id, so path-keyed chunks would collapse onto the same nodes and the cutover would delete chunks the pending document just wrote. Keying on `uid` is correct. The narrow fix is to have `IngestionPipeline.run` derive a stable uid when the caller supplies no `document_info`, mirroring `GraphRAG._resolve_document_id` (`main.py:1491`) — that leaves `update()` untouched, since it always passes an explicit pending id.

### 384 default

The PR's own sentence is the finding: *"Twice the text got 1.3% more facts out — the relationship extractor returns a roughly constant number of facts per call, which is the real ceiling."* That's a **prompt ceiling**, not a chunk-size result. Smaller chunks don't lift it; they buy more facts by buying more calls.

Measured on a 53k-token corpus with the actual chunker:

| `max_tokens` | chunks / LLM calls | input tokens | tok/chunk |
|---|---|---|---|
| 384 | 241 | 180,759 | 306 |
| 512 | 179 | 151,746 | 404 |
| 768 | 97 | — | 668 |

- ❌ **512 → 384 is +35% LLM calls and +19% input tokens**, every ingest, for every user, against a reported gain of relation F1 0.237 vs 0.223. The fixed per-call prompt overhead is 444 tokens, so 107k of those 181k input tokens are the same instructions re-sent 241 times. That's quality bought at a 35% call-count premium — a legitimate trade, but it should be stated as one.
- ❌ **The measurement is likely stale.** #310 rewrites the relation prompt specifically to lift the constant-facts-per-call ceiling (+15.5% reply length). If facts-per-call rises, the optimal chunk size moves back up with it. This number should be re-run on top of #310's prompt before it becomes the shipped default.
- ⚠️ **Metric mismatch for a global default.** Entity/relation F1 measure graph construction; #310 states repeatedly that entity-quality wins moved the final answer score zero times. Chunk size is also the retrieval unit — 306-token chunks mean smaller retrieved context and more fragmented citations, which wasn't measured.
- ❌ **Docs and CHANGELOG not updated.** The `src/` sweep is complete (no `512` left), but `docs/strategies.mdx:136,151,180` still say "default: 512" and `docs/benchmark.mdx:77` calls 512 the SDK default.

---

### Suggested

1. Split the PR — the ID fix can land as soon as the pipeline gap and a re-ingest test are in.
2. Fix the direct-`IngestionPipeline` uid derivation; add one end-to-end re-ingest test.
3. Add the upgrade note about the one-time extra chunk copy.
4. Hold 384 until it's re-measured on top of #310. If it lands regardless: CHANGELOG entry naming the +35% call-count cost, and update the three docs references.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -181,6 +215,8 @@ async def run(
                 ctx.log("No chunks produced, pipeline complete")
                 return IngestionResult(document_info=document.document_info)
 
+            _assign_deterministic_chunk_uids(document.document_info, chunks)
```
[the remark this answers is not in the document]
Naseem77: Done in d38cf36: the content hash + `get_document_record()` check now runs right after load, before step 2, and returns `chunks_indexed=0`. Added `test_unchanged_document_never_reaches_the_chunker` (asserts `chunk_document` and `index_chunks` are not called on the skip path). Also moved the `content_hash` write to after steps 8/9 so a failed extraction is retried rather than skipped (`test_content_hash_is_written_only_after_a_successful_run`).

Naseem77:
Thanks — every point checked against the code and addressed in d38cf36.

**Chunk-id fix**
- *Only takes effect through `GraphRAG.ingest()`* — confirmed: loaders leave `DocumentInfo.uid` at `uuid4()`. The pipeline now derives the id itself when no `document_info` is passed: `os.path.normpath(source)` in file mode, a text hash in text mode — the same rule as `_resolve_document_id`, keyed on uid (not path) exactly for the `update()` reason you gave. Pending ids are still passed explicitly and untouched.
- *No end-to-end test* — added `TestReingestIdempotencyEndToEnd`: two `pipeline.run("report.txt")` calls with no `document_info` → identical chunk-id sets and the same Document id; same text in two different files → disjoint ids; text mode stable too.
- *Upgrade side effect* and *edited-file re-ids* — both in the CHANGELOG and in the helper's docstring.
- *Dead `uid=sc.uid`* — removed.

**384 default**
- The +35 % calls / +19 % input tokens is real — our own full-pipeline runs show 103 → 157 extraction calls — and it is now stated in the CHANGELOG and PR body rather than implied.
- On "re-run on top of #310's prompt": done, 3 repeats per arm on the full stack (#309 + #310 + #317) against main with its own `LLMVerifiedResolution` switched on, same day, same machine. Exact relation F1 0.062 → 0.134 (2.2×), entity F1 0.550 → 0.573, entities 0.74 → 0.94 × gold, relation vocabulary 316 → 101 labels, **answer accuracy on 148 questions 27.0 → 32.0 %**, QA input tokens −18 %, total cost +$0.03 per run (+9 %). So the facts-per-call ceiling did move with #310's prompt and 384 still wins on the metric that matters. I kept the two changes in one PR on that basis; happy to split if you still prefer.
- Retrieval unit: QA is the retrieval measurement we have and it went up while retrieved tokens went down; citation fragmentation per se was not measured.
- Docs updated (`strategies.mdx` ×3, `benchmark.mdx`).

Also from the bot reviews: the unchanged-document check now runs before chunking with `chunks_indexed=0`, and `content_hash` is written only after a successful run so a failed extraction is retried. Full suite 1112 passed / 41 skipped; live re-ingest probe on FalkorDB: 0 provider calls.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -153,9 +201,34 @@ async def run(
                     document_info=document_info or DocumentInfo(path=source),
                 )
                 ctx.log("Using provided text (loader skipped)")
+                if document_info is None:
+                    # Text mode with no caller-supplied identity: derive the
+                    # Document id from the text so the same text ingested
+                    # twice is the same document (and the chunk ids below,
+                    # which include the document id, are stable too).
```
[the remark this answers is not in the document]
Naseem77: Done in 5ae8cdd: `GraphRAG._resolve_document_id` now derives `text-<16hex>` from a SHA-256 of the text, so `ingest(text=...)` without `document_id` is stable and matches the pipeline; explicit `document_id` still wins. Test added (`test_text_mode_document_id_is_derived_from_the_text`).

CHANGELOG.md
```diff
@@ -9,6 +9,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- Re-ingesting the same file no longer duplicates its chunks. Chunk ids are
+  now derived from document id + position + text instead of a fresh
+  `uuid4()`, so `MERGE` finds the existing node (17 → 17 → 17 chunks across
+  three ingests, previously 17 → 34 → 51). `ContextualChunking` hashes the
+  original chunk text, not the LLM-enriched one. Applies to
+  `IngestionPipeline.run()` directly as well as through `GraphRAG.ingest()`:
+  when no `document_info` is supplied the pipeline now derives a stable
+  Document id from the normalised source path (or, in text mode, from the
+  text). **Upgrade note:** graphs built before this change hold random chunk
```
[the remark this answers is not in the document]
Naseem77: Went the other way in 5ae8cdd — the code now matches the claim: the facade hashes the text too, and the CHANGELOG spells out that `GraphRAG.ingest(text=...)` without `document_id` is one document per distinct text (pass `document_id` to store identical text twice).

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -173,6 +246,46 @@ async def run(
...
+            # changed ontology or strategy with the same text is update()'s
+            # job.
+            doc_uid = document.document_info.uid
+            if doc_uid:
+                existing = await self.graph_store.get_document_record(doc_uid)
+                if existing is not None and existing.content_hash == content_hash:
+                    ctx.log(
+                        f"Document '{doc_uid}' already ingested with identical content; "
+                        f"skipping. Call update() to re-extract."
+                    )
```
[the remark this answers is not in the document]
Naseem77: Fixed in 5ae8cdd — the message now names `GraphRAG.update()` explicitly.

CHANGELOG.md
```diff
@@ -9,6 +9,39 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Changed
+
+- Default chunk size lowered from 512 to 384 tokens in
+  `SentenceTokenCapChunking`, `StructuralChunking`, `ContextualChunking` and
+  the documented `CallableChunking` example. Measured on the benchmark corpus:
+  entity F1 0.574 vs 0.563 and relation F1 0.237 vs 0.223 against 768; with
+  the current extraction prompt, exact relation F1 2.2× and answer accuracy
+  27 → 32 % for the full stack. **Cost:** ~35 % more extraction LLM calls and
+  ~19 % more input tokens per ingest (103 → 157 calls on the 11-document
+  corpus), since the per-call instructions are re-sent once per chunk. Pass
```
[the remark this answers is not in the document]
Naseem77: Correct — 103 → 157 is +52 %. The +35 % came from the reviewer's 53k-token corpus (179 → 241). CHANGELOG now gives both figures with their corpora (5ae8cdd).

docs/benchmark.mdx
```diff
@@ -74,7 +74,7 @@ model and the judge.
 | Generation temperature | 0.7 | Appendix H.2 |
 | Framework | GraphRAG-SDK 1.3.0 (PyPI) on FalkorDB | — |
 | Graph layout | one graph per corpus document | — |
-| Chunking | `SentenceTokenCapChunking`, max_tokens 512, overlap 2 sentences | SDK default |
+| Chunking | `SentenceTokenCapChunking`, max_tokens 384 (was 512 before 1.2), overlap 2 sentences | SDK default |
```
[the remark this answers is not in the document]
Naseem77: Changed to "was 512 before this release" (5ae8cdd).

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -153,9 +201,34 @@ async def run(
...
+                # the same file was a new Document — and, since the chunk
+                # ids include the document id, a new set of chunks. Derive
+                # the id from the normalised source path, the same rule
+                # ``GraphRAG._resolve_document_id`` applies, so the
+                # pipeline is idempotent on its own and not only through
+                # the facade. ``update()`` always passes an explicit
+                # pending id and is unaffected.
+                if document_info is None:
+                    document.document_info = DocumentInfo(
+                        uid=os.path.normpath(source),
```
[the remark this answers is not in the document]
Naseem77: Valid — and `GraphRAG._resolve_document_id` had the same `normpath` on main. 5ae8cdd adds `core.models.stable_document_id(source)`: filesystem paths are normalised, anything with a URI scheme (`https://`, `s3://`, `file://` …) is returned verbatim, and both the pipeline and the facade use it. Tests cover `https://example.test/doc?path=a/../b` staying distinct from `https://example.test/b`, plus the pipeline using the URI verbatim as the Document id.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -271,6 +374,13 @@ async def _step_index_chunks() -> None:
                 _step_index_chunks(),
             )
 
+            # Step 9b: only now record the content hash. Writing it in step 3
+            # meant a failure in extraction or the graph write left a
+            # Document that looked complete, and the next ingest skipped it
+            # (Copilot review on #309). A partial run therefore retries in
+            # full; only a finished run is recognised as unchanged.
+            await self._mark_content_hash(document.document_info.uid, content_hash)
```
galshubeli: **`content_hash` is stamped even when the writes it certifies partially failed.**

Moving the hash write to step 9b fixes the case where the run *raises* — but two of the write paths never raise:

- `GraphStore.upsert_relationships` (`graph_store.py:279-287`) catches per-relationship exceptions, logs a warning, and continues, returning a count.
- `VectorStore.index_chunks` returns 0 both when no embedder is configured and when every embedding call fails.

So a transient FalkorDB error or an embedding rate-limit leaves some `RELATES` edges unwritten and/or chunks with no `embedding` property, `run()` returns normally, the hash gets stamped — and every later `ingest()` of that file short-circuits at the new check. The missing edges and vectors are then never repaired and the document silently drops out of chunk vector search.

This is strictly worse than pre-PR behaviour, where the next re-ingest would have healed it. Suggest only writing the hash when the writes actually reported success (e.g. have `upsert_relationships`/`index_chunks` surface a partial-failure signal, and skip `_mark_content_hash` on it).

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -181,22 +295,11 @@ async def run(
                 ctx.log("No chunks produced, pipeline complete")
                 return IngestionResult(document_info=document.document_info)
 
+            _assign_deterministic_chunk_uids(document.document_info, chunks)
```
galshubeli, line 360: **The deterministic-uid invariant doesn't survive `update()`.**

`_assign_deterministic_chunk_uids` hashes `doc_info.uid`, but `update()` runs the pipeline under a randomised id — `api/main.py:2173`:

```python
pending_id = f"{resolved_id}__pending__{uuid4().hex[:8]}"
```

The Phase-5 cutover renames the Document to `resolved_id` but leaves the chunk ids as digests of that random pending id. So for any document that has been through `update()`, "chunk id is reproducible from document id + position + text" no longer holds.

Concrete consequence: for such a document — or any pre-upgrade document with no stored `content_hash`, where the skip can't fire — a later full `ingest()` of the same content computes `H(resolved_id, i, text)`, matches none of the stored chunks, and duplicates the entire chunk layer. That's exactly the failure fix #1 is meant to eliminate.

Deriving the uid from the *final* resolved id rather than the pending one would close this.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -173,6 +246,47 @@ async def run(
...
+            # Same rule update() uses for its no-op: a stored Document whose
+            # content_hash matches means every chunk, entity and edge is
+            # already there — the hash is only written once a run has
+            # completed, so a half-finished ingest is never mistaken for a
+            # finished one. Changed text (new hash) takes the full path; a
+            # changed ontology or strategy with the same text is update()'s
+            # job.
+            doc_uid = document.document_info.uid
+            if doc_uid:
+                existing = await self.graph_store.get_document_record(doc_uid)
```
galshubeli: **Unguarded `get_document_record` breaks the duck-typed graph-store contract.**

`self.graph_store` is declared `graph_store: Any` at line 136 with an explicit "import avoided for layering" note, and before this change `IngestionPipeline` only ever called `upsert_nodes` / `upsert_relationships` on it. This adds a hard `get_document_record` dependency on the very first step of every run.

Any user-supplied graph-store adapter or test double that implements the previously-required surface now raises `AttributeError`, wrapped as `IngestionError("Pipeline failed: ...")`, for every ingest.

```python
get_record = getattr(self.graph_store, "get_document_record", None)
if get_record is not None:
    existing = await get_record(doc_uid)
    ...
```

Skipping the check when the method is absent keeps the contract intact and degrades gracefully to pre-PR behaviour.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -173,6 +246,47 @@ async def run(
...
+                if existing is not None and existing.content_hash == content_hash:
+                    ctx.log(
+                        f"Document '{doc_uid}' already ingested with identical content; "
```
galshubeli, line 342: **This message points users at `update()`, which no-ops on the same condition.**

`update()` short-circuits on exactly this comparison at `api/main.py:2154`:

```python
if existing.content_hash == new_hash:
    ctx.log(f"update: content hash matches for '{resolved_id}', no-op")
    return UpdateResult(..., no_op=True, replaced_existing=True)
```

...returning before Phase 2. So after this PR there's no supported way to re-extract byte-identical content under a changed ontology, chunker or extractor: `ingest(path, chunker=X)` silently returns zeros, and `update(path)` returns `no_op=True`. The user has to delete the document first.

Given the 384-token default lands in this same PR, re-chunking existing corpora is a fairly likely thing for people to want. Either the guidance needs correcting, or both paths need a `force=` / `if_unchanged=` escape hatch.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -1369,10 +1370,13 @@ async def ingest(
 
         Uses sensible defaults for any unspecified strategy:
```
galshubeli, line 1385: **The public `ingest()` docstring wasn't updated alongside the private `_resolve_document_id` one.**

`_resolve_document_id` below got a thorough rewrite, but this docstring still documents the old id semantics — text mode "defaults to a generated `text-<8hex>` id" (now a 16-hex SHA-256 of the text) and file mode "defaults to `os.path.normpath(source)`" (URI sources are now kept verbatim) — and says nothing about the new unchanged-document no-op. `update()` at line ~2034 repeats the stale normpath claim.

Worth calling out because it hides a breaking change on the primary public API: a caller that today does `ingest(text=...)` per record and relies on each call producing a distinct Document now silently gets one Document plus a zero-result no-op for every duplicate text. That deserves an explicit note in the docstring and probably in the CHANGELOG's breaking-changes section too.

galshubeli (commented):
Reviewed the full diff hunk by hunk plus the surrounding `pipeline.py`, `api/main.py` (ingest/update/phase-0 recovery), `storage/graph_store.py`, `storage/vector_store.py` and the chunking strategies. Ran the suite against this branch — all green, matching your numbers.

The measurement work behind all three fixes is solid and the 384 justification is convincing. Six findings below; **two look merge-blocking**, both in the same area:

1. **`content_hash` is stamped on partially-failed runs** (`pipeline.py:382`) — `upsert_relationships` and `index_chunks` swallow per-item failures, so a rate-limited or flaky ingest gets permanently marked complete and can never self-heal.
2. **`update()` breaks the deterministic-uid invariant** (`pipeline.py:298`) — the pipeline runs under a random `__pending__` id, so chunk ids from an update aren't reproducible, and a later `ingest()` duplicates the whole chunk layer.

Findings 1, 3 and 4 are arguably facets of one design gap: the content hash is treated as proof that a complete, current extraction exists, but nothing enforces that it's only written when that's true, and nothing lets you override it when the extraction *parameters* rather than the content have changed.

---

One finding has no inline anchor because the file isn't in this PR's diff:

**`docs/graphrag-accuracy-benchmark.mdx:106`** still documents `max_tokens 512` as the "SDK default" while `docs/benchmark.mdx:90` was updated to 384. The file arrived via the merge from `main`, so it's an easy one to miss, but as shipped the two docs state contradictory defaults.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
galshubeli: Fixed in `dedae86` — `getattr(self.graph_store, "get_document_record", None)` with the fallback spelled out in the comment. Duck-typed stores now just never short-circuit, which is the pre-PR behaviour. Thanks.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
galshubeli: Fixed in `dedae86` — `update(force=True)` added and threaded through `update_sync`, and this message now names it. Good escape hatch.

graphrag_sdk/src/graphrag_sdk/api/main.py
galshubeli: Fixed in `dedae86` — both `ingest()` and `update()` now document the normalised-path id, `text-<16hex>`, the `__pending__` restriction and the unchanged no-op, and the CHANGELOG has an explicit breaking-change note for text-mode dedup.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
galshubeli: Taking the documented-limitation route is fair, and the CHANGELOG note is clear about why pending and live can't share chunk nodes during cutover.

One caveat: the argument rests on *"the `content_hash` the cutover records is what keeps a later `ingest()` of that content a no-op"* — i.e. correctness here depends on the hash always being present. Two things in `dedae86` weaken that (both in the new review): `update()` doesn't consult `incomplete_writes` before stamping, and no-embedder setups never record a hash at all. In the no-embedder case a document that's been through `update()` has mismatched chunk ids *and* no hash, so the next `ingest()` duplicates its whole chunk layer — the scenario this note says can't happen.

Not asking you to fix the pending-id scheme. Just noting the limitation is only safe once the hash is reliable again.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
galshubeli: The `_reported_short` + `incomplete_writes` mechanism is the right shape, and treating non-int returns as trustworthy is a nice touch — it keeps duck-typed stores and mocks working instead of refusing to ever mark them complete.

Two gaps left, both raised as inline comments on the new review:

- `update()` never consults `pipeline_result.metadata["incomplete_writes"]`; Phase 5 stamps `content_hash=new_hash` unconditionally, so the same partial-write corruption is still reachable, just via `update()` rather than `ingest()`.
- `index_chunks` returns 0 when no embedder is configured, which now reads as a permanent shortfall — those deployments never record a hash and never get the skip.

The `ingest()` path itself looks correct.

galshubeli (commented):
Re-reviewed at `dedae86`. Suite is green against the PR source — **1146 passed, 41 skipped**.

Four of the six are cleanly closed — replies on those threads. The `_reported_short` / `incomplete_writes` design for the hash gate is the right shape, and `update(force=True)` is a good escape hatch.

**Still blocking, though**, and the first one is the original finding rather than a new one:

1. **`update()` bypasses the complete-writes gate** — Phase 5 stamps `content_hash=new_hash` unconditionally and never reads `incomplete_writes`. The partial-write corruption is still reachable, just through `update()` instead of `ingest()`.
2. **No-embedder deployments never record a hash** — `index_chunks` returns 0 with no embedder configured, which now reads as a permanent shortfall, so the unchanged-document skip never fires for graph-only setups.
3. **`graphrag_sdk/.venv` is committed** as a symlink into `/Users/naseemali/...`.

2 and 3 are small. 1 is the real work.

On the `__pending__` chunk-id limitation (F2): documenting rather than fixing is a reasonable call and I'm not asking you to change the scheme — but the CHANGELOG's argument depends on `content_hash` always being present, which is exactly what 1 and 2 undermine. Details on that thread.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -2144,11 +2177,18 @@ async def update(
             )
 
         if existing.content_hash == new_hash:
```
galshubeli, line 2189: **`update()` bypasses the new complete-writes gate.**

`run()` now correctly withholds `content_hash` when a write reports a shortfall — but `update()` doesn't inherit that. The pipeline result is captured at line 2238 and `incomplete_writes` is only ever read for the return payload (line 2308); it is never checked before Phase 5:

```python
chunks_deleted = await self._graph_store.rollforward_cutover(
    pending_id=pending_id,
    real_id=resolved_id,
    path=doc_path,
    content_hash=new_hash,      # unconditional
)
```

So a dropped `RELATES` edge or an unembedded chunk during an `update()` still gets certified complete, and the document is skipped forever after. That's the original finding, reachable through a different entry point — so this one isn't closed yet.

Simplest fix: check `pipeline_result.metadata.get("incomplete_writes")` before Phase 5 and either pass `content_hash=None` through the cutover or fail the update outright. Passing `None` keeps the cutover's crash-safety intact while leaving the document eligible for repair.

graphrag_sdk/src/graphrag_sdk/ingestion/pipeline.py
```diff
@@ -258,31 +433,60 @@ async def run(
...
-            async def _step_mentions() -> int:
+            async def _step_mentions() -> tuple[int, str | None]:
                 ctx.log("Step 8/9: Writing mentions (uncapped)")
                 return await self._write_mentions(graph_data, ctx)
 
-            async def _step_index_chunks() -> None:
+            async def _step_index_chunks() -> str | None:
                 ctx.log("Step 9/9: Embedding & indexing chunks")
-                await self.vector_store.index_chunks(chunks)
+                indexed = await self.vector_store.index_chunks(chunks)
```
galshubeli, line 450: **A missing embedder is counted as an incomplete write, which disables the skip entirely.**

`VectorStore.index_chunks` returns 0 when no embedder is configured (`vector_store.py:196-198`):

```python
if not self._embedder:
    logger.warning("No embedder configured — skipping chunk indexing")
    return 0
```

That's indistinguishable here from "every embedding failed", so `_reported_short(0, len(chunks.chunks))` is true on **every** ingest for a graph-only deployment. Consequences: `content_hash` is never recorded, the unchanged-document no-op never fires, and each ingest logs a warning about incomplete writes that the user can't act on. Fix #3 — the headline feature of this PR — is silently off for anyone running without embeddings.

The pipeline can't tell the two cases apart from the return value alone. Cleanest fix is at the source: return `None` from `index_chunks` when there's no embedder, which `_reported_short` already treats as "take it at its word". A guard in `_step_index_chunks` would work too, but the pipeline has no visibility into embedder config, so the vector store is the right place.

Worth a test for the no-embedder path — the current suite doesn't cover it.

graphrag_sdk/.venv
```diff
@@ -0,0 +1 @@
+/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
```
galshubeli: **This shouldn't be committed** — it's a symlink to an absolute path on your machine:

```
/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
```

It's dangling for every other checkout, and it broke my first test run (imports silently resolved against the wrong tree). Please `git rm --cached graphrag_sdk/.venv` and add `.venv` to `.gitignore`.

galshubeli (approved):

Issue #286, opened 2026-07-22
docs(ingestion): concurrency tuning guide + warning for the sequential per-file ingest() pattern


galshubeli:
# Ingestion concurrency: better defaults, tuning guide, and a warning for the per-file ingest() loop pattern

## Problem

Two related pitfalls make first-run ingestion look much slower than the SDK actually is:

1. **Conservative defaults.** `ingest()` batches at `max_concurrency=3` documents in flight (`src/graphrag_sdk/api/main.py:1579`) and the extractor defaults to 12 concurrent LLM calls (`ingestion/extraction_strategies/graph_extraction.py:523`). Reasonable for laptops against rate-limited APIs; far below what a local model server or enterprise API tier can sustain. Nothing tells the user these knobs exist.

2. **Silent zero-parallelism pattern.** Calling `ingest(source)` in a loop over files — the most natural-looking usage — gives zero document-level parallelism, with no warning. Passing the list in one call is what engages `_ingest_batch`'s semaphore. A POC evaluating the SDK this way concluded "PDF ingestion is a bottleneck" when it was running one document at a time.

## Proposed improvements

- **Docs: a tuning guide** — one page covering `max_concurrency`, extractor `max_concurrency`, `update_concurrency` (see #284), which stage dominates (LLM extraction for typical PDF workloads, not the graph write), and how to tell (per-step ctx logs already exist in `pipeline.run()`).
- **Runtime hint** — detect the sequential-single-source pattern (e.g. N single-source `ingest()` calls within one process/session) and log an INFO/WARNING pointing at list-based ingestion.
- **Revisit defaults** — consider raising the document default and/or deriving it from the model provider (local endpoints tolerate much more than metered APIs).

Related: #282/#283 fix the write path; this issue fixes the perception gap where extraction parallelism is left on the table.

Issue #284, opened 2026-07-22
perf(update): make concurrent update() safe by design — update_concurrency=1 serializes incremental refreshes


galshubeli:
# Make concurrent update() safe so update_concurrency can be raised above 1

## Problem

`update()` defaults to `update_concurrency=1`, and the docstring is explicit: *"Default is 1, and you almost certainly should not raise it"* (`src/graphrag_sdk/api/main.py:2439`) — orphan cleanup is only correct under concurrent updates because of a fragile ordering invariant (MENTIONED_IN edges persisted before `run()` returns; see comments in `ingestion/pipeline.py:241-260`). Raising it risks wrongly deleting a shared entity when two concurrent updates both see it as an orphan candidate.

For update-heavy workloads this is the structural bottleneck: a production weekly refresh of ~18K modified documents runs fully serialized — **~55-hour ETA (≈3 days) for a weekly job**. Even with #282 and #283 landed, document-level serialization caps throughput at one document's pipeline latency per document.

## Fix direction

Make concurrent updates safe **by design** rather than by ordering invariant, so the default can be raised. Candidate approaches:

- **Entity-level locking** — serialize only updates that touch the same entity ids.
- **Partition-by-shared-entity scheduling** — pre-compute candidate entity overlap between modified docs; run non-overlapping partitions concurrently.
- **Harden the existing invariant** — make orphan-candidate checks transactional/recheck-before-delete so a concurrent writer cannot invalidate them between check and delete.

Whatever the approach, it must be designed against **`IncrementalResolution`** (the default resolver as of #278) — its graph-aware linking reads the live graph during resolution, so concurrency safety for orphan cleanup and for resolution reads should be solved together, not separately.

## Acceptance

- `update_concurrency > 1` documented as safe (and ideally a higher default).
- A concurrency stress test covering the shared-entity orphan race (extend `test_concurrent_updates_preserve_shared_entity`).

Related: #282, #283 (per-query costs; this issue is the parallelism ceiling above them).

Issue #283, opened 2026-07-22
perf(ingestion): bulk write path for relationships — avoid per-doc per-type sequential UNWIND round trips


galshubeli:
# Bulk write path for relationships — per-doc per-type sequential round trips dominate ingestion

## Problem

Writes happen per document inside `pipeline.run()` (`src/graphrag_sdk/ingestion/pipeline.py:232-269`), and `upsert_relationships` groups by relationship type (`storage/graph_store.py:157-162`). Consequences:

- A document touching 25-30 relationship types pays **25-30 sequential UNWIND round trips**, each nearly empty.
- The 500-row `_BATCH_SIZE` batching effectively never engages — per-doc per-type groups are far smaller than 500.
- Per-label node upserts, relationship upserts, and mentions writes are all sequential `await`s.

Measured in a production incremental-update workload (~18K docs): **~20 min per 100-doc batch, dominated by ~25-30 UNWIND queries × network latency**. An SDK user asked directly: *"Does the GraphRAG SDK have a bulk write path for edge insertion that avoids per-type-per-batch round trips?"* — today the answer is no.

## Fix — two stages

**Stage 1 (cheap, high value): concurrent flushes.** Fire the per-type / per-label UNWIND batches with `asyncio.gather` instead of awaiting sequentially. Directly attacks the round-trips × latency term. Note FalkorDB serializes writes per graph, so this hides *network* latency rather than server work — the server-side cost is addressed by #282.

**Stage 2 (opt-in): cross-document write buffer.** Accumulate nodes/edges across documents and flush per type at full batch size (e.g. a batch-session context manager with explicit flush). Must be **opt-in**: today each document is atomic-ish with its own `IngestionResult`, and a shared buffer changes failure attribution — a silent behavior change would break per-doc error handling.

## Related failure-mode multiplier

The per-item fallback (`graph_store.py:232-249`) degrades one failing 500-row batch into 500 sequential queries. At production scale this turns a transient batch error into minutes of extra latency. Worth bounding or logging loudly as part of this work.

Related: #282 (range indexes — the server-side half of the same slowdown).

Issue #282, opened 2026-07-22
perf(storage): auto-create range indexes on id — writes degrade to label scans as the graph grows


galshubeli:
# Auto-create range indexes on `id` — writes currently degrade to label scans

## Problem

`ensure_indices()` (`src/graphrag_sdk/storage/vector_store.py:588`) creates only **vector** and **fulltext** indexes. No range (exact-match) index is ever created on the `id` property, yet the entire write path keys on it:

- **Node upserts** — `MERGE (n:`Label` {id: item.id})` (`storage/graph_store.py:108`): label scan per UNWIND row.
- **Relationship upserts** — `MATCH (a:`Src` {id}), (b:`Tgt` {id})` (`storage/graph_store.py:212-227`): **two** label scans per edge row, typically against `__Entity__` or `Chunk`.
- **Mentions writes** (`_write_mentions`) — same pattern.
- **Update/delete machinery** — `delete_document` commit point, `has_pending_delete`, `delete_stale_relationships`, orphan cleanup: all `MATCH ... {id: $id}`.

Cost is quadratic over the life of the graph: each batch scans a graph made bigger by all previous batches. At a production deployment's scale (~290K nodes / 1.2M edges), one 500-row edge batch performs up to 1,000 endpoint lookups × ~137K-node label scans ≈ **~137M node visits for a single UNWIND query**. This is the dominant server-side write cost and the main reason ingestion "gets slower the more you load."

Reported from two independent production/POC deployments.

## Fix

1. Add range indexes to `ensure_indices()`:
   - `CREATE INDEX FOR (n:__Entity__) ON (n.id)`
   - `CREATE INDEX FOR (n:Chunk) ON (n.id)`
   - `CREATE INDEX FOR (n:Document) ON (n.id)`

   FalkorDB's `CREATE INDEX` errors when the index exists; the existing `_INDEX_EXISTS_MARKERS` handling (`vector_store.py:27,112`) already makes this idempotent — reuse it.

[...]

PR #310, opened 2026-08-25
Extraction: entity and relation quality, memory and throughput fixes

Naseem77:
## Extraction — entities and relations

Every claim below was measured on an 11-document benchmark corpus against a real
FalkorDB, not estimated. Part of the ingestion investigation in FalkorDB/research#88.

18 commits, one per fix, each verified by running the code before it was committed.

---

# Entities

### Updated / fixed

- **The name-finding model default changed** — `urchade/gliner_medium-v2.1` →
  `knowledgator/gliner-bi-small-v2.0`. Ceiling recall 0.805 vs 0.709, 432MB vs 781MB on
  disk, 1004MB vs 1532MB resident, ~14s vs ~29s, 2048-word window vs 384. Smaller,
  faster, finds more.
- **The confidence cutoff was one hardcoded number applied to every model** — now a
  per-model table with a sane fallback and a warning, so swapping models can't silently
  mis-tune the threshold. That single number was the actual cause of a 4-second spaCy
  model appearing to beat our shipped config: the bi-encoder models return **2 entities
  for an entire corpus** at the 0.75 that suits the old model, and raise nothing.
- **The name-finder silently ignored everything past word 388 of a chunk.** No exception,
  no warning — the tail never reached the model. It now slides overlapping windows across
  the whole chunk and merges the results; without it recall drops 0.805 → 0.621. Spans
  are re-offset into the original text so downstream span code stays correct.
- **Specific dates like "January 5, 1992" were stored as things in the graph** — 63 of 274
  false entities (23%). Now rejected by name. Precision 0.577 → 0.642, recall unchanged.
  Periods ("the 1820s", "the Abbasid era") deliberately stay.
- **`US` the country was thrown away as `us` the pronoun** — the stoplist check casefolded
  first. Short all-caps tokens are now read as acronyms.
- **Junk tokens (shell fragments, pronouns, operators) were only filtered by asking the
  LLM nicely** — they're now a hard rule in code as a backstop. A prompt instruction can
  be quietly skipped; a list cannot.
- **Every upload loaded its own copy of the model** (~395MB each, ~11.6GB for 30
  documents) — one shared process-wide cache now, ~1.39GB.
- **A lock forced name-finding one chunk at a time despite the model being safe in
  parallel** — inference now takes no lock at all, **1.56×** faster. (#71 claimed 3.44×;
  the honest measured figure is 1.56×, because torch already uses the cores.) Thread
  safety was tested, not assumed: 40/40 documents byte-identical to the serialised run.
- **New `SpacyExtractor`** — a second, different name-finder, added and exported.
- **New `CompositeExtractor`** — runs several finders and merges overlapping spans, added
  and exported.
- **An entity type named `Document` collided with our own bookkeeping and vanished from
  search** — 106 documents reported in an 11-document corpus, *and* the node never got
  `__Entity__` so it silently dropped out of dedup and retrieval. Now rejected up front on
  both the `entity_types` and ontology paths, and the spaCy mapping that caused it is fixed.
- **Ingest never said when chunks failed, so a half-empty graph passed as a clean upload**
  — it now reports which chunks failed, by id, and warns rather than logging "0 nodes".

### Dropped — measured, and the evidence went the other way

- **A bigger name-finding model.** The large one finds fewer names at 2.3× the memory.
- **Replacing GLiNER's family.** Five architectures tested; ours has the highest recall.
- **Keeping "unsure" names (`candidate_threshold`).** In the code, off by default. Lowering
  the threshold outright raised recall 32% but dropped entity F1 0.568 → 0.474 and triple
  F1 0.236 → 0.211.
- **`CompositeExtractor` as the default.** Built and exported, but the default stays plain
  GLiNER — combining extractors didn't move the end-to-end score.
- **Re-tuning thread fan-out.** The "24 threads" number doesn't exist in our code.
- **Entity quality as a direction.** Six wins here, and the final answer score moved zero
  times.

---

# Relations

### Updated / fixed

- **We invented our own private vocabulary** — 447 distinct relation labels against 30 in
  the reference, 68.2% used exactly once, only 17.3% of edges carrying a label gold also
  uses, and `contains` (123 gold triples) emitted **zero** times. We now ship
  `DEFAULT_RELATION_TYPES`, exported. Any fixed list doubles exact triple F1
  (0.0725 → 0.1490); a list written *without* looking at gold scored as well as gold's own
  vocabulary (0.1490 vs 0.1456) — the gain is consistency, not word choice.
  `relation_types=[]` stays meaningful for anyone who wants the old free-for-all.
- **The relationship step stopped early because it decided it had said enough** — ~12
  relations per chunk using 2.5k of a 16k budget. A rewritten prompt fixed it.
- **Writing to the graph got 477× slower as it grew because there was no index** — one is
  now created automatically per label, including `__Entity__` and relation endpoints.
  50K nodes: 111.47s → 0.69s total, and flat per batch instead of 4.6ms → 2194.8ms.

### Dropped — measured, and the evidence went the other way

- **Splitting the one overloaded LLM call in two.** Worse. The design we'd called our top
  bug was the winner.
- **The dedicated entity verifier.** Fully built (`VERIFY_ENTITIES_PROMPT`,
  `_verify_entities`, quote-checking so a broken judge can't empty your graph) and shipped
  **off by default**, because its deletions were as good as random: 57 correct / 41 junk
  removed vs 56.6 / 41.4 expected by chance. It answers "is this in the text" correctly;
  the question we need answered is *salience*, which the text cannot answer.
- **Deleting the "check the entities" instruction.** 94 junk entities poured in
  (719 → 813, precision 0.645 → 0.551). Kept, with a comment saying so.
- **A free local model instead of the paid one.** Not close.
- **Asking the model a second time.** Better graph, no better answers. Code deleted.
- **Type rules.** Shipped for years, never measured; they only help a crowded room.
- **Grab-everything-then-filter.** Cutting 55% of relationships made the graph 47% better
  and the answers worse.
- **The hidden 5-second timeout.** Doesn't reproduce — a 10.6s query returned fine.
- **Throwing on extraction failure.** One bad chunk would kill an entire bulk upload.

---

### Verification

Full suite on this branch: **1153 passed, 41 skipped**. Runtime also dropped from ~146s to
~18s, which is the shared model cache and the smaller default model showing up in CI.

Refs: FalkorDB/research#88, FalkorDB/research#71, #282, #283, #284, #286

## Summary by CodeRabbit

- **New Features**
  - Added spaCy and combined entity extraction options.
  - Improved long-text extraction, confidence handling, model reuse, and noise/date filtering.
  - Added default relationship types and optional entity verification.
  - Added chunk failure reporting and clearer extraction status tracking.
  - Improved graph storage with automatic label-specific indexing.
  - Exposed additional extraction tools and relation settings through the public API.

- **Bug Fixes**
  - Prevented reserved structural labels from being used as entity types.
  - Continued graph writes when index creation fails.

- **Documentation**
  - Clarified confidence-band behavior for entity extraction.

### DIFF
--- a/docs/extraction.mdx
+++ b/docs/extraction.mdx
@@ -81,7 +81,8 @@ A local transformer model that runs on your machine — no API calls needed.
 from graphrag_sdk import GLiNERExtractor
 
 extractor = GLiNERExtractor(
-    threshold=0.75,                          # confidence threshold (below = "Unknown")
+    threshold=0.75,                          # at or above: typed entity
+    candidate_threshold=0.5625,              # band below threshold kept as "Unknown" (default: 25 % below threshold)
     model_name="urchade/gliner_medium-v2.1", # HuggingFace model
 )
 ```
@@ -90,7 +91,7 @@ extractor = GLiNERExtractor(
 1. The model is loaded lazily on first use (thread-safe via `threading.Lock`)
 2. Runs inference via `asyncio.to_thread()` to avoid blocking the event loop
 3. Returns predictions with character-level spans — more precise than LLM spans
-4. Entities below the confidence threshold are labeled `"Unknown"` (not discarded)
+4. Entities in the band just below the confidence threshold (default: 25 % below it, e.g. 0.5625–0.75) are labeled `"Unknown"` and handed to the step-2 LLM, which re-types or drops them against the text; anything below the band is discarded. Pass `candidate_threshold=None` to disable the band.
 
 **Best for:** Production pipelines where you want fast, cheap NER without API costs.
 
@@ -198,7 +199,7 @@ Person, Organization, Technology, Product, Location,
 Date, Event, Concept, Law, Dataset, Method
 ```
 
-Entities that don't match any type (or fall below the confidence threshold) are labeled `"Unknown"`.
+Entities that don't match any type (or score in the band just below the confidence threshold) are labeled `"Unknown"`.
 
 ### Customizing the Ontology
 
--- a/graphrag_sdk/src/graphrag_sdk/__init__.py
+++ b/graphrag_sdk/src/graphrag_sdk/__init__.py
@@ -91,11 +91,14 @@ from graphrag_sdk.ingestion.extraction_strategies.coref_resolvers import (
     FastCorefResolver,
 )
 from graphrag_sdk.ingestion.extraction_strategies.entity_extractors import (
+    CompositeExtractor,
     EntityExtractor,
     GLiNERExtractor,
     LLMExtractor,
+    SpacyExtractor,
 )
 from graphrag_sdk.ingestion.extraction_strategies.graph_extraction import (
+    DEFAULT_RELATION_TYPES,
     GraphExtraction,
 )
 from graphrag_sdk.ingestion.loaders.base import LoaderStrategy
@@ -130,6 +133,7 @@ from graphrag_sdk.storage.ontology_store import (
 from graphrag_sdk.storage.vector_store import VectorStore
 
 __all__ = [
+    "DEFAULT_RELATION_TYPES",
     # Version
     "__version__",
     # API
@@ -191,8 +195,10 @@ __all__ = [
     "CachedChunkExtraction",
     "GraphExtraction",
     "EntityExtractor",
+    "CompositeExtractor",
     "GLiNERExtractor",
     "LLMExtractor",
+    "SpacyExtractor",
     "CorefResolver",
     "FastCorefResolver",
     "IngestionPipeline",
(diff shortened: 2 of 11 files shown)


Naseem77:
Superseded by five stacked PRs, one topic each:

- #318 GLiNER perf + failed-chunk reporting
- #319 entity-name hygiene (dates, acronyms, junk, reserved labels)
- #320 windowing, per-model thresholds, "Unknown" band
- #321 relation vocabulary + no early stop
- #322 SpacyExtractor / CompositeExtractor (opt-in)

Review in order #318 → #322; each retargets to `main` as the previous one merges.

PR #318, opened 2026-09-07
Extraction: share one GLiNER model, drop the inference lock, index ids, report failed chunks

Naseem77:
## Extraction — GLiNER speed & memory, and failed chunks reported

Measured on an 11-document benchmark against a real FalkorDB. Part of FalkorDB/research#88. First of five stacked PRs that replace #310.

### 1. One GLiNER model per process
**Problem:** every document being ingested loaded its own copy of the GLiNER model — 6 documents in one call = 6 models = 2.4 GB; 30 documents ≈ 11.6 GB.
**Fix:** a class-level cache keyed by model name; one copy shared by every document (+0 MB per document; 30 documents = 1.4 GB).

### 2. Inference lock removed
**Problem:** a lock around `predict_entities` serialised the chunks of a document even though they were dispatched to threads.
**Fix:** lock removed. GLiNER inference mutates no model state (verified: 8 documents in parallel vs serial, 5 trials, 40/40 identical). 1.56× faster NER.

### 3. Index on `id` per entity label
**Problem:** `MERGE` on `id` with no index scans every node — writes went from 4.6 ms to 2,195 ms per 500 entities as the graph grew to 50k nodes (477×).
**Fix:** `GraphStore` creates a range index on `id` for each label once, before its first write. Flat: 8.8 → 8.5 ms; 50k nodes 111 s → 0.69 s.

### 4. Failed chunks are reported
**Problem:** a chunk whose LLM call failed (timeout, bad JSON) contributed nothing and `ingest()` still reported success — an empty document and a broken run looked identical.
**Fix:** `GraphData` / `IngestionResult` carry `chunks_attempted` and `failed_chunks` (the ids, retryable) and a warning is logged.

### Verification
Full suite: **1126 passed, 41 skipped**; ruff check + format clean.

## Summary by CodeRabbit

* **New Features**
  * Extraction reports now include attempted, skipped, entity-failed, and relationship-failed chunks.
  * Ingestion results expose detailed extraction reports for monitoring and targeted retries.
  * Graph storage automatically prepares ID indexes for graph updates.
  * Document and chunk identifiers are normalized consistently for repeatable ingestion.

* **Improvements**
  * Shared extraction models reduce duplicate loading and improve concurrent processing.
  * Relationship failures preserve successfully extracted entities.
  * Extraction reports remain available after filtering, pruning, and cached extraction.
  * Unchanged documents can skip redundant processing.

* **Bug Fixes**
  * Failed or incomplete writes are retried instead of being treated as complete.

### DIFF
--- a/graphrag_sdk/src/graphrag_sdk/core/models.py
+++ b/graphrag_sdk/src/graphrag_sdk/core/models.py
@@ -555,13 +555,39 @@ class Ontology(DataModel):
 
 
 class GraphData(DataModel):
-    """Entities and relationships extracted from text."""
+    """Entities and relationships extracted from text.
+
+    ``chunks_attempted`` / ``failed_chunks`` exist because a per-chunk
+    extraction failure is otherwise invisible: failures are swallowed and the
+    chunk contributes nothing, so a document where every call failed returns
+    exactly what a document containing no entities returns — same type, same
+    empty lists, no exception. Callers had no way to tell "nothing to find"
+    from "found nothing because everything broke", and a *partial* failure
+    silently shipped a half-empty graph that looked successful.
+
+    Read them together: ``0`` entities from ``0`` attempted chunks is an empty
+    document; ``0`` from ``14`` is a broken one. ``failed_chunks`` carries the
+    chunk uids that raised so the caller can retry just those (see
+    ``BackfillExecutor``, which follows the same convention) rather than
+    re-ingesting the document.
+    """
 
     nodes: list[GraphNode] = Field(default_factory=list)
     relationships: list[GraphRelationship] = Field(default_factory=list)
     mentions: list[EntityMention] = Field(default_factory=list)
     extracted_entities: list[ExtractedEntity] = Field(default_factory=list)
     extracted_relations: list[ExtractedRelation] = Field(default_factory=list)
+    chunks_attempted: int = 0
+    failed_chunks: list[str] = Field(default_factory=list)
+
+    @property
+    def extraction_failed(self) -> bool:
+        """True when every attempted chunk failed.
+
+        Distinguishes total extraction failure from a genuinely empty
+        document, which reports ``chunks_attempted == 0``.
+        """
+        return self.chunks_attempted > 0 and len(self.failed_chunks) == self.chunks_attempted
 
 
 class ExtractedEntity(DataModel):
--- a/graphrag_sdk/src/graphrag_sdk/core/models.py
+++ b/graphrag_sdk/src/graphrag_sdk/core/models.py
@@ -557,17 +557,29 @@ class Ontology(DataModel):
 class GraphData(DataModel):
     """Entities and relationships extracted from text.
 
-    ``chunks_attempted`` / ``failed_chunks`` exist because a per-chunk
-    extraction failure is otherwise invisible: failures are swallowed and the
-    chunk contributes nothing, so a document where every call failed returns
-    exactly what a document containing no entities returns — same type, same
-    empty lists, no exception. Callers had no way to tell "nothing to find"
-    from "found nothing because everything broke", and a *partial* failure
-    silently shipped a half-empty graph that looked successful.
+    The report fields exist because a per-chunk extraction failure is
+    otherwise invisible: failures are swallowed, so a document where every
+    call failed returns exactly what a document containing no entities
+    returns — same type, same empty lists, no exception. Callers had no way
+    to tell "nothing to find" from "found nothing because everything broke",
+    and a *partial* failure silently shipped a half-empty graph that looked
+    successful.
 
-    Read them together: ``0`` entities from ``0`` attempted chunks is an empty
-    document; ``0`` from ``14`` is a broken one. ``failed_chunks`` carries the
-    chunk uids that raised so the caller can retry just those (see
+    - ``chunks_attempted``: chunks that were sent to extraction.
+    - ``chunks_skipped``: chunks never attempted because the latency budget
+      ran out first. A truncated run is *not* a healthy short one; this is
+      how the caller tells them apart.
+    - ``failed_chunks``: uids whose entity extraction (step 1) raised. These
+      chunks produced no entities of their own.
+    - ``relation_failed_chunks``: uids whose entity extraction succeeded but
+      whose relationship extraction (step 2) failed. Their entities are in
+      the graph; their edges are not. Disjoint from ``failed_chunks``.
+
+    Read the lists, not the entity counts: a successful extraction can
+    legitimately yield zero entities (``chunks_attempted > 0``,
+    ``failed_chunks == []``), and a run with every chunk in
+    ``relation_failed_chunks`` has a complete node set and no edges. Both
+    lists carry chunk uids so the caller can retry just those (see
     ``BackfillExecutor``, which follows the same convention) rather than
     re-ingesting the document.
     """
@@ -578,14 +590,20 @@ class GraphData(DataModel):
     extracted_entities: list[ExtractedEntity] = Field(default_factory=list)
     extracted_relations: list[ExtractedRelation] = Field(default_factory=list)
     chunks_attempted: int = 0
+    chunks_skipped: int = 0
     failed_chunks: list[str] = Field(default_factory=list)
+    relation_failed_chunks: list[str] = Field(default_factory=list)
 
     @property
     def extraction_failed(self) -> bool:
-        """True when every attempted chunk failed.
+        """True when entity extraction failed for every attempted chunk.
 
-        Distinguishes total extraction failure from a genuinely empty
-        document, which reports ``chunks_attempted == 0``.
+        This is the "no chunk produced entities" signal. Step-2 (relationship)
+        failures do not count: those chunks still contributed their nodes, so
+        a caller that aborts or retries on this flag would otherwise discard a
+        good entity extraction. Check ``relation_failed_chunks`` for lost
+        edges. A genuinely empty document reports ``chunks_attempted == 0``
+        and is not a failure.
         """
         return self.chunks_attempted > 0 and len(self.failed_chunks) == self.chunks_attempted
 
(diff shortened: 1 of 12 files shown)


graphrag_sdk/src/graphrag_sdk/core/models.py
```diff
@@ -555,13 +555,39 @@ def _merge_props(existing: list[Attribute], incoming: list[Attribute]) -> list[A
...
+    re-ingesting the document.
+    """
 
     nodes: list[GraphNode] = Field(default_factory=list)
     relationships: list[GraphRelationship] = Field(default_factory=list)
     mentions: list[EntityMention] = Field(default_factory=list)
     extracted_entities: list[ExtractedEntity] = Field(default_factory=list)
     extracted_relations: list[ExtractedRelation] = Field(default_factory=list)
+    chunks_attempted: int = 0
+    failed_chunks: list[str] = Field(default_factory=list)
```
galshubeli, line 638: These two fields never reach a caller.

`Pipeline.run()` calls `extract()` (`pipeline.py:203`), then `_filter_quality` (206) and `_prune` (210) — both of which rebuild `GraphData` field-by-field:

```python
# pipeline.py:486
new_gd = GraphData(
    nodes=valid_nodes,
    relationships=valid_rels,
    mentions=graph_data.mentions,
    extracted_entities=graph_data.extracted_entities,
    extracted_relations=graph_data.extracted_relations,
)
```

`chunks_attempted` and `failed_chunks` aren't in the list, so they fall back to their defaults. Every real ingest therefore ends with `chunks_attempted == 0`, `failed_chunks == []`, and `extraction_failed` permanently `False`. `CachedChunkExtraction._merge` (`cached_chunk_extraction.py:373`) does the same, so `update()` discards the report before the pipeline ever sees it.

`_remap_mentions` (`pipeline.py:528`) already does this the safe way — `graph_data.model_copy(update={"mentions": rewritten})` — which stays correct as fields are added. Worth using in all three places.

Separately, `IngestionResult` (`models.py:731`) is untouched, so even with the above fixed there's no field on the pipeline's return value carrying the report out.

The new tests pass because they call `GraphExtraction.extract()` directly; nothing exercises the path through the pipeline.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -537,6 +541,7 @@ async def _step1(text: str, chunk_uid: str) -> list[ExtractedEntity]:
                         f"Step 1 NER failed for chunk {active_chunks[i].index}: {result}",
                         logging.WARNING,
                     )
+                    failed_chunk_uids.add(active_chunks[i].uid)
```
galshubeli, line 556: This is the only place a step-1 failure gets recorded, and it's the local-extractor branch. The `LLMExtractor` branch above appends `[]` for both `not item.ok` (506-511) and `item.response is None` (512-517) without ever calling `failed_chunk_uids.add(chunk.uid)`.

So on the default LLM path: chunk 3's NER call times out → `chunk_entities[3] = []` → step 2 runs for that chunk with an empty `entities_json`, usually returns fine → the chunk is reported clean. That's the silent-timeout case the `GraphData` docstring describes, still open for the extractor most callers run.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -568,6 +573,7 @@ async def _step1(text: str, chunk_uid: str) -> list[ExtractedEntity]:
 
         all_entities: list[ExtractedEntity] = []
         all_relations: list[ExtractedRelation] = []
+        rels_by_chunk: dict[int, list[ExtractedRelation]] = {}
```
galshubeli: `rels_by_chunk` doesn't change the output. `chunk_idx` is unique per iteration, so `setdefault(...).extend(rels)` at 614 creates a one-element bucket each time, and the flatten at 616-617 walks them in insertion order — the same sequence the previous inline `all_relations.extend(rels)` produced. Nothing groups, dedupes, or reads the per-chunk structure.

Either drop it and extend `all_relations` directly at 614, or make it earn its place — per-chunk relation counts would pair well with `failed_chunks`.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -580,6 +586,10 @@ async def _step1(text: str, chunk_uid: str) -> list[ExtractedEntity]:
                         f"Step 2 verify+rels failed for chunk {chunk.index}: {item.error}",
                         logging.WARNING,
                     )
+                    # Relations for this chunk are lost even though step 1
+                    # entities survive, so it counts as failed: the caller
+                    # needs to know this chunk's edges were never extracted.
+                    failed_chunk_uids.add(chunk.uid)
```
galshubeli: This conflicts with what `extraction_failed` promises.

If every chunk's step-1 NER succeeds and every step-2 call fails, `all_entities` still receives the step-1 entities at 594 and the nodes are written — but `failed_chunks == chunks_attempted`, so `extraction_failed` returns `True` for a run that produced a complete node set. The property's docstring says "True when every attempted chunk failed", and `GraphData`'s says "`0` from `14` is a broken one". A caller that aborts or retries on `extraction_failed` would discard a good entity extraction.

The comment is right that the caller needs to know the edges were lost — but that's a different fact from "extraction failed". A separate relation-failure list, or excluding relation-only failures from the `extraction_failed` predicate, keeps both signals honest.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -629,12 +642,22 @@ async def _step1(text: str, chunk_uid: str) -> list[ExtractedEntity]:
             mentions=all_mentions,
             extracted_entities=merged_entities,
             extracted_relations=merged_relations,
+            chunks_attempted=len(active_chunks),
```
galshubeli, line 655: `len(active_chunks)` is the post-truncation count. `active_chunks` is cut short by the `ctx.budget_exceeded` guard at 460-465, so a 40-chunk document that runs out of budget after 3 chunks reports `chunks_attempted=3, failed_chunks=[], extraction_failed=False` — indistinguishable from a fully extracted 3-chunk document.

That's the "partial failure silently shipped a half-empty graph that looked successful" case the docstring says these fields close. Using `len(chunks.chunks)` as the denominator, or adding an explicit skipped/truncated count, would surface it. Same gap at the early return on 468, which returns a bare `GraphData(nodes=[], relationships=[])`.

graphrag_sdk/src/graphrag_sdk/storage/graph_store.py
```diff
@@ -60,6 +61,45 @@ def __init__(self, connection: FalkorDBConnection) -> None:
...
+        results are memoised per label. The cache is per-store-instance, which
+        is the right scope: a new store means a possibly different graph.
+
+        Failures are logged and swallowed. A missing index is a performance
+        problem; a raised exception here would be a correctness problem, and
+        the caller's write must not depend on the optimisation succeeding.
+        """
+        if safe_label in self._indexed_labels:
+            return
+        self._indexed_labels.add(safe_label)
```
galshubeli: The label is marked indexed before the index exists, and line 100 swallows the failure — so one transient error permanently disables indexing for that label.

FalkorDB restarts during the first write of a run → `_conn.query` exhausts its retries and raises `DatabaseUnavailableError` → logged at debug and dropped, but `safe_label` was already added at 97. `GraphStore` is a process-lifetime singleton (`api/main.py:292`), so the index is never attempted again and the 477x degradation documented in this docstring quietly comes back for the rest of the process.

Moving the `add()` inside the `try`, after a successful `query`, keeps the "failures are swallowed" contract while letting the next batch retry.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -330,28 +330,89 @@ def __init__(
         self._threshold = threshold
         self._model_name = model_name
         self._model: Any = None
+        # Retained only so callers/tests that swap in a context manager to
+        # A/B the removed inference lock keep working. Nothing in this class
+        # acquires it any more; model loading uses the class-level
+        # ``_CACHE_LOCK`` and inference deliberately takes no lock at all.
+        # See ``_predict_sync`` for the thread-safety evidence.
         self._lock = threading.Lock()
```
galshubeli: Two things around this line.

**The class docstring is now wrong.** Line 316 still says the model is "loaded lazily on first use and protected by a lock so a single instance can be safely shared across concurrent `asyncio.to_thread` calls". After this PR inference takes no lock, and loading is guarded by the class-level `_CACHE_LOCK`, not `self._lock`. It's the public statement of the concurrency contract, so it should move with the change.

**`self._lock` and the `_predict_body` split (407) are scaffolding for a harness that isn't in the tree.** The comment keeps them for "callers/tests that swap in a context manager to A/B the removed inference lock", but nothing does — `test_entity_extractors.py:213` asserts `self._lock` is absent from `_predict_sync` rather than exercising the hook. If the benchmark isn't landing with the PR, both can go; the measurement is already preserved in the comment at 388-404.

galshubeli (commented):
Reviewed the four source files against `main`.

The two performance changes hold up. `GLiNER.inference` is `@torch.no_grad()`-decorated and assigns no instance state beyond `self.eval()`, and `data_processor` mutates state only in `__init__` — no evidence of a race, so dropping the inference lock looks safe. On the index side, `_NON_TRANSIENT_MARKERS` (`core/connection.py:240`) already covers `already indexed` / `already exists`, so a duplicate-index error raises immediately instead of burning the retry budget.

The failed-chunk reporting needs work. The fields are computed correctly inside `GraphExtraction.extract()`, but they never survive to a caller, and three semantic gaps mean the signal doesn't yet mean what the docstring says. Details inline. Two findings land in files this diff doesn't touch, so they're here instead:

**`pipeline.py` and `cached_chunk_extraction.py` drop the new fields** — see the inline note on `core/models.py`. This one makes the feature a no-op outside the new unit tests, so it's the one I'd fix first.

**`GraphStore.delete_all()` doesn't reset `_indexed_labels`** (`storage/graph_store.py:439`). `_conn.delete_graph()` issues `GRAPH.DELETE`, which drops the graph key and every index with it, but the store still believes each previously seen label is indexed. `delete_all()` followed by a re-ingest on the same instance runs the whole rebuild unindexed — exactly the case this PR exists to fix. `self._indexed_labels.clear()` in `delete_all()` covers it (harmless on the `DETACH DELETE` fallback path too).

galshubeli (approved):

PR #319, opened 2026-09-07
Extraction: dates, acronyms, junk tokens and reserved labels — entity-name hygiene

Naseem77:
## Extraction — entity-name hygiene

Part of FalkorDB/research#88. Stacked on #318. Four cases where a non-entity became a graph node, or a real entity was thrown away, all fixed in code (`is_valid_entity_name` / `_reject_reserved_labels`) rather than by asking the LLM.

### 1. Specific dates are not entities
**Problem:** "1896", "14 January 1904", "1003 CE" became nodes (23 % of the junk that reached the graph).
**Fix:** `is_specific_date()` rejects them as entity names when the ontology has no `Date` type; periods ("1820s", "19th century") are kept, and trailing sentence punctuation ("1823.") does not get a date past the gate. **Precondition:** `DEFAULT_ENTITY_TYPES` declares `Date`, so on the out-of-the-box path the rule never fires and the numbers above (23 % of junk, precision 0.577 → 0.642) apply only to a custom ontology that omits `Date`. Extractors receive the label list without an ontology (`entity_types=None` means "no date rule"); `GraphExtraction._step1` applies the gate once the ontology is known, so grounded discovery's NER anchor labels do not drop "1984" or "747".

### 2. `US`, `UN`, `EU` were dropped as pronouns
**Problem:** the stop-word filter lower-cased first, so `US` matched `us`.
**Fix:** a 2–3 letter all-caps token is read as an acronym and is exempted from the *pronoun* rows of the stop-list only; lower-case `us` is still removed, and `CD`, `ETC`, `MAN` in an all-caps heading are still rejected as the shell tokens and generic nouns they are. Residual: other all-caps pronouns (`HE`, `WE`, `MY`) pass, since they cannot be told apart from `US`/`IT` without context.

### 3. Junk tokens filtered in code
**Problem:** `+=`, `->`, `cd`, `he`, `they` reached the graph whenever the LLM ignored the prompt line asking it to drop them.
**Fix:** a hard rule in `is_valid_entity_name` (operators, shell abbreviations, generic 1–2 char tokens), in addition to the prompt line. Tests now assert behaviour, not prompt wording.

### 4. `Document` and `Chunk` rejected as entity types
**Problem:** the store uses those labels for its own nodes; an entity typed `Document` was stored as a file, invisible to dedup and retrieval (11 files uploaded → graph reported 106 documents).
**Fix:** `RESERVED_NODE_LABELS = {"Document", "Chunk"}`, the single definition shared by `GraphStore`, the Cypher generator and `reject_reserved_labels`. The match is **exact** (like the store's own membership test and FalkorDB labels), so `document` is allowed. The primary guard is in `OntologyStore.register`, before anything is persisted; discovery's `_validate_proposal` returns it as a retryable error so the LLM can pick another label; `backfill_entity` and `GraphExtraction.__init__`/`extract()` keep the check as defence-in-depth. The message names the clashing label.

### Verification
Full suite: **1210 passed, 41 skipped**; `ruff check src` and `ruff format --check src` clean. Note CI does not run on this PR (`ci.yml` triggers on `main`/`staging` only), so the stack's merger should run it on 3.10/3.12.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reserved labels such as `Document` and `Chunk` from being used as custom entity types.
  * Improved entity extraction by filtering noise, invalid names, and unsupported dates according to the configured ontology.
  * Ensured backfilled entities use consistent identifiers and connect correctly to existing mentions.
  * Discovery now reports invalid reserved labels as validation errors, allowing correction without aborting.
  * Prevented invalid ontology registrations and backfills from writing partial data.

### DIFF
--- a/graphrag_sdk/.venv
+++ b/graphrag_sdk/.venv
@@ -1,0 +1,1 @@
+/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
--- a/graphrag_sdk/.venv
+++ b/graphrag_sdk/.venv
@@ -1,1 +1,0 @@
-/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -38,6 +38,7 @@ from graphrag_sdk.core.models import (
     RagResult,
     RetrieverResult,
     UpdateResult,
+    reject_reserved_labels,
 )
 from graphrag_sdk.core.providers import Embedder, LLMInterface
 from graphrag_sdk.discovery import SchemaExtensionProposal, suggest_extensions
@@ -1019,6 +1020,9 @@ class GraphRAG:
         nothing else.
         """
         await self._ensure_ontology_initialized()
+        # A reserved label that reached the ontology (e.g. persisted before the
+        # register-time guard existed) must not become real ``:Document`` nodes.
+        reject_reserved_labels([label])
         entity = next(
             (e for e in self._global_ontology.entities if e.label == label),
             None,
@@ -1059,6 +1063,10 @@ class GraphRAG:
             or "(none)"
         )
         target_desc_line = f"- description: {entity.description}\n" if entity.description else ""
+        # The date gate keys off the whole ontology, not the label being filled:
+        # backfilling ``Product`` in a graph that declares ``Date`` must keep
+        # "747" exactly as the normal extraction path does.
+        ontology_labels = [e.label for e in self._global_ontology.entities]
 
         def _prompt(ctx: ChunkContext) -> str:
             return BACKFILL_ENTITY_PROMPT.format(
@@ -1083,10 +1091,12 @@ class GraphRAG:
             new_mentions: list[GraphRelationship] = []
             for ent in parsed:
                 ent_name = (ent.get("name") or "").strip()
-                if not is_valid_entity_name(ent_name, [label]):
+                if not is_valid_entity_name(ent_name, ontology_labels):
                     skipped += 1
                     continue
-                ent_id = compute_entity_id(label, ent_name)
+                # (name, type) -- the same id the extraction path assigns, so a
+                # backfilled entity merges with its extracted twin.
+                ent_id = compute_entity_id(ent_name, label)
                 props: dict[str, Any] = {
                     "name": ent_name,
                     "description": ent.get("description", ""),
--- a/graphrag_sdk/src/graphrag_sdk/api/main.py
+++ b/graphrag_sdk/src/graphrag_sdk/api/main.py
@@ -1083,7 +1083,7 @@ class GraphRAG:
             new_mentions: list[GraphRelationship] = []
             for ent in parsed:
                 ent_name = (ent.get("name") or "").strip()
-                if not is_valid_entity_name(ent_name):
+                if not is_valid_entity_name(ent_name, [label]):
                     skipped += 1
                     continue
                 ent_id = compute_entity_id(label, ent_name)
--- a/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
+++ b/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
@@ -176,7 +176,10 @@ def is_valid_entity_name(name: str) -> bool:
     stripped = name.strip()
     if len(stripped) < MIN_NAME_LEN or len(stripped) > MAX_NAME_LEN:
         return False
-    if stripped.lower() in _ENTITY_STOPLIST:
+    # "US" is a country, "us" is a pronoun, and casefolding the stoplist check
+    # conflates them. An all-caps short token is an acronym, not a pronoun.
+    is_acronym = len(stripped) <= 3 and stripped.isupper() and stripped.isalpha()
+    if not is_acronym and stripped.lower() in _ENTITY_STOPLIST:
         return False
     if is_specific_date(stripped):
         return False
--- a/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
+++ b/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
@@ -67,7 +67,19 @@ _PRONOUNS: set[str] = {
     "its",
 }
 
-_ENTITY_STOPLIST: set[str] = _PRONOUNS | {
+# Shell and system abbreviations that read as entities to an NER model but name
+# nothing in any document domain.  Ported here from the extraction prompt, which
+# used to ask the LLM to remove them: a fixed list costs nothing, cannot vary
+# between runs, and cannot be quietly skipped the way the prompt instruction was
+# (see RESULTS.md P2.10).  Well-known two-letter acronyms - AI, US, UK, EU, UN,
+# Go - are deliberately absent and stay.
+_SHELL_TOKENS: frozenset[str] = frozenset({
+    "sh", "cd", "ls", "rm", "cp", "mv", "dt", "bg", "fg", "fn", "df", "du",
+    "ps", "cat", "pwd", "echo", "mkdir", "rmdir", "chmod", "chown", "grep",
+    "awk", "sed", "env", "sudo", "ssh", "tmp", "var", "usr", "bin", "etc",
+})
+
+_ENTITY_STOPLIST: set[str] = _PRONOUNS | _SHELL_TOKENS | {
     # Generic/anonymous references
     "narrator",
     "the narrator",
@@ -182,6 +194,10 @@ def is_valid_entity_name(name: str) -> bool:
     if not is_acronym and stripped.lower() in _ENTITY_STOPLIST:
         return False
     if is_specific_date(stripped):
+        return False
+    # Operator and punctuation tokens (+=, ->, ==, !=). A name with no letter or
+    # digit anywhere in it cannot be the name of anything.
+    if not any(ch.isalnum() for ch in stripped):
         return False
     return True
 
--- a/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
+++ b/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
@@ -127,6 +127,48 @@ def _normalize_type_label(raw: str) -> str:
     return re.sub(r"[\s_\-/]+", "", s)
 
 
+# A date that pins down a single moment is an *attribute* of an event ("the
+# lamp was installed in 1823"), not something you can hold a conversation
+# about.  A date that names a *period* ("the 1820s", "the Abbasid era") is a
+# thing facts get attached to, so it stays.
+#
+# Measured on the 11-document benchmark: specific dates were 63 of 274 false
+# positive entities (23%).  Removing them lifted entity precision 0.577 -> 0.642
+# with recall unchanged at 0.644 (F1 0.609 -> 0.643) and cost nothing, because
+# every gold Date entity is a decade.  This also stops the graph accumulating
+# "X happened_in 1957" edges that carry no answerable content.
+_SPECIFIC_DATE_RE = re.compile(
+    r"""^(?:
+        (?:c\.?\s*|circa\s+|ca\.?\s*)?\d{3,4}\s*(?:ce|bce|ad|bc)?   # 1823, 1003 ce, c. 1200
+      | \d{1,2}\s+[a-z]+\s+\d{4}                                     # 14 january 1904
+      | [a-z]+\s+\d{1,2},?\s+\d{4}                                    # january 14, 1904
+      | \d{4}[-/]\d{1,2}(?:[-/]\d{1,2})?                              # 1904-01-14
+    )$""",
+    re.IGNORECASE | re.VERBOSE,
+)
+
+# Overrides the rule above: these name a span of time, not a moment.
+_DATE_PERIOD_RE = re.compile(
+    r"\d{3,4}s\b|centur|era\b|dynasty|period|decade|age\b", re.IGNORECASE
+)
+
+
+def is_specific_date(name: str) -> bool:
+    """True if name pins down one moment in time rather than naming a period.
+
+    Specific dates are rejected as entity names: they belong on the relation
+    that mentions them, not as nodes of their own.  Note the deliberate gap --
+    a product genuinely named for a number ("747", "1984") is indistinguishable
+    from a year here and will be dropped.  That trade was worth 63 false
+    positives against zero true positives on the benchmark corpus, but it is
+    the first thing to revisit if a domain uses numeric product names.
+    """
+    stripped = name.strip()
+    if _DATE_PERIOD_RE.search(stripped):
+        return False
+    return bool(_SPECIFIC_DATE_RE.match(stripped))
+
+
 def is_valid_entity_name(name: str) -> bool:
     """Return True if name passes quality gates for entity extraction."""
     if not name or not name.strip():
@@ -135,6 +177,8 @@ def is_valid_entity_name(name: str) -> bool:
     if len(stripped) < MIN_NAME_LEN or len(stripped) > MAX_NAME_LEN:
         return False
     if stripped.lower() in _ENTITY_STOPLIST:
+        return False
+    if is_specific_date(stripped):
         return False
     return True
 
--- a/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
+++ b/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
@@ -200,8 +200,16 @@ _SPECIFIC_DATE_RE = re.compile(
     re.IGNORECASE | re.VERBOSE,
 )
 
-# Overrides the rule above: these name a span of time, not a moment.
-_DATE_PERIOD_RE = re.compile(r"\d{3,4}s\b|centur|era\b|dynasty|period|decade|age\b", re.IGNORECASE)
+# Overrides the rule above: these name a span of time, not a moment.  "era"
+# and "age" are whole words so "opera" and "package" are not periods.
+_DATE_PERIOD_RE = re.compile(
+    r"\d{3,4}s\b|centur|\bera\b|dynasty|period|decade|\bage\b", re.IGNORECASE
+)
+
+# Punctuation that extraction can leave on a date at a sentence boundary
+# ("1823.", "(14 January 1904)"); stripped before the date regexes run so it
+# cannot smuggle a date past the gate.
+_DATE_EDGE_PUNCTUATION = ".,;:!?()[]{}\"'"
 
 
 def is_specific_date(name: str) -> bool:
@@ -216,7 +224,7 @@ def is_specific_date(name: str) -> bool:
     positives against zero true positives on the benchmark corpus, but it is
     the first thing to revisit if a domain uses numeric product names.
     """
-    stripped = name.strip()
+    stripped = name.strip().strip(_DATE_EDGE_PUNCTUATION)
     if _DATE_PERIOD_RE.search(stripped):
         return False
     return bool(_SPECIFIC_DATE_RE.match(stripped))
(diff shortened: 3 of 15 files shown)


graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -127,14 +173,78 @@ def _normalize_type_label(raw: str) -> str:
...
+    """
     if not name or not name.strip():
         return False
     stripped = name.strip()
     if len(stripped) < MIN_NAME_LEN or len(stripped) > MAX_NAME_LEN:
         return False
-    if stripped.lower() in _ENTITY_STOPLIST:
+    # "US" is a country, "us" is a pronoun, and casefolding the stoplist check
+    # conflates them. An all-caps short token is an acronym, not a pronoun.
+    is_acronym = len(stripped) <= 3 and stripped.isupper() and stripped.isalpha()
```
galshubeli, line 262: **High — the acronym exemption bypasses the *entire* stoplist, not just pronouns.**

`is_acronym` gates the whole `_ENTITY_STOPLIST` check on the next line, but the comment above only justifies exempting pronouns (`us` vs `US`). Executing the module confirms the blast radius — every one of these returns `True`:

- shell tokens: `CD, LS, RM, PS, CAT, ENV, ETC, VAR, BIN, USR, TMP, SSH, SED, AWK`
- pronouns: `HE, HER, HIS, ITS, MY, WE`
- generic nouns: `ONE, MAN, BOY`

So the new `_SHELL_TOKENS` filter — this PR's headline fix — is fully defeated for uppercase input. A technical doc with all-caps headings/table cells, or OCR'd text, puts `CD`, `ETC`, `MAN`, `HE` straight into the graph as entities. The tests only exercise lowercase (`["sh","cd","ls",...]`), so this stays green.

Suggest narrowing the exemption to the pronoun subset only, e.g. check the non-pronoun stoplist unconditionally and only skip `_PRONOUNS` when `is_acronym`.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -127,14 +173,78 @@ def _normalize_type_label(raw: str) -> str:
...
+    if not is_acronym and stripped.lower() in _ENTITY_STOPLIST:
+        return False
+    if not _ontology_has_date_type(entity_types) and is_specific_date(stripped):
```
galshubeli: **Medium — grounded discovery passes NER *anchor* labels into this gate, so numeric-looking mentions are dropped before catalog linking.**

`discovery/pipeline.py:589` calls `extractor.extract_entities(chunk.text, _NER_ANCHOR_LABELS, chunk.uid)`, where `_NER_ANCHOR_LABELS = ["person", "organization", "location", "event"]` — documented at `pipeline.py:569` as "deliberately broad and never appear in the output ontology". `_parse_predictions` now forwards that list here as if it were the ontology, so `_ontology_has_date_type` is `False` and every date-shaped or bare 3–4 digit mention is discarded.

Verified: with the anchor list, `is_valid_entity_name("1984")`, `("747")`, `("1969")`, `("1776")` are all `False`; all four are `True` under `DEFAULT_ENTITY_TYPES`.

Concrete impact: discovering an ontology over a corpus about the novel *1984* or the Boeing *747* — those mentions never reach `catalog.link_entity`, so `Book`/`Aircraft` never land in `detected_types` and never enter the discovered ontology. The anchor list isn't an ontology and shouldn't drive this gate.

galshubeli: **Low — the date filter is inert under the default configuration.**

`DEFAULT_ENTITY_TYPES` (line 31) contains `"Date"`, and `api/main.py:418` seeds exactly those labels into the ontology store when no ontology is supplied. So on the out-of-the-box path `_ontology_has_date_type` is `True` and `is_specific_date` never rejects anything.

The PR body advertises "23% of the junk" and precision 0.577 -> 0.642, but no default-configuration user gets that — it only fires for a custom ontology that omits `Date`. Either drop `Date` from `DEFAULT_ENTITY_TYPES`, or state the precondition in the PR description so the numbers aren't read as the default-path gain.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -336,6 +337,38 @@ def _optional_extras(obj: Any) -> dict[str, Any]:
...
+    2. ``GraphStore._write_nodes`` marks a node as an entity only when its
+       label is *not* structural, so the node never receives ``__Entity__``
+       and is dropped from deduplication and retrieval. It is written, then
+       ignored.
+
+    Neither failure raises, so the graph just quietly degrades. Fail here
+    instead, at configuration time, where the caller can act on it.
+    """
+    reserved = {label.casefold(): label for label in RESERVED_NODE_LABELS}
+    clashes = [t for t in types if str(t).strip().casefold() in reserved]
```
galshubeli: **Low — this rejection is case-insensitive but the store's membership test is case-sensitive, so previously-safe configs now fail.**

`clashes` casefolds, while `GraphStore._write_nodes` (`storage/graph_store.py:127`) tests `label not in self._STRUCTURAL_LABELS` exactly. A lowercase `"document"` entity type used to behave correctly — it received `__Entity__`, and since FalkorDB labels are case-sensitive `MATCH (d:Document)` never picked it up — yet `GraphExtraction(entity_types=["document"])` now raises.

Either narrow this check to exact matches, or casefold the store's structural test too so the two agree on what "reserved" means. (Minor: `reserved` on the line above is built as a dict whose values are never used — a set would do.)

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -442,7 +475,10 @@ async def extract(
     ) -> GraphData:
         # Resolve entity types: ontology overrides instance default
         if ontology.entities:
-            entity_types = [e.label for e in ontology.entities]
+            # Ontology labels bypass the constructor, so re-check here: an
+            # ontology declaring a reserved label must fail as loudly as
+            # passing one to entity_types would.
+            entity_types = _reject_reserved_labels([e.label for e in ontology.entities])
```
galshubeli, line 461: **Medium — for the ontology path this fires mid-pipeline, not "at configuration time", and nothing guards the persistence path.**

The docstring above `_reject_reserved_labels` says failure happens "at configuration time, where the caller can act on it". True for `__init__`, but here it runs inside `extract()`, which is step 4/9 of `ingestion/pipeline.py:203` — *after* step 3 has already written the `Document` node, all `Chunk` nodes and the `PART_OF`/`NEXT_CHUNK` chain. The `ValueError` is wrapped into `IngestionError` at `pipeline.py:297`, leaving a half-ingested document behind.

Worse, the bad label can be *persisted* before it's ever checked:
- `api/main.py:589` (`add_entity`) registers an arbitrary `Entity(label=...)` with no reserved-label check.
- `discovery/pipeline.py`'s `_validate_proposal` validates reserved *attribute* names but not entity labels.

So an LLM-proposed or user-added `Document` type gets stored in the ontology graph, and then *every* subsequent `ingest()` fails hard at step 4, orphaning Document/Chunk nodes each time. The check would be better placed in `_ontology_store.register` / `_validate_proposal` — the latter has a retry loop that would let the LLM correct itself.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -1083,7 +1083,7 @@ async def _merge(parsed: list[dict[str, Any]], ctx: ChunkContext) -> BackfillMer
             new_mentions: list[GraphRelationship] = []
             for ent in parsed:
                 ent_name = (ent.get("name") or "").strip()
-                if not is_valid_entity_name(ent_name):
+                if not is_valid_entity_name(ent_name, [label]):
```
galshubeli: **Medium — the Date gate is given a single label instead of the ontology's labels.**

Passing `[label]` makes date-keeping depend on whether the label *currently being backfilled* is `Date`. Every other `is_valid_entity_name` call site passes the full allowed-type list, and `self._global_ontology.entities` is in scope right here.

Scenario: a graph whose ontology declares `Date`, running `backfill_entity("Product")` over a catalogue with numeric product names ("747", "1984") — those are silently dropped and only surface in `values_skipped`, while the identical names survive the normal extraction path.

Suggest `is_valid_entity_name(ent_name, [e.label for e in self._global_ontology.entities])`.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -1083,7 +1083,7 @@ async def _merge(parsed: list[dict[str, Any]], ctx: ChunkContext) -> BackfillMer
...
                     skipped += 1
                     continue
                 ent_id = compute_entity_id(label, ent_name)
```
galshubeli: **High — `compute_entity_id` arguments are swapped here** (pre-existing, but one line below a line this PR edits, and it's the same bug class the PR is fixing).

The signature is `compute_entity_id(name, entity_type="")`, and every other call site passes `(ent.name, ent.type)` — see `graph_extraction.py:662`, `:948`, `:1007-1008`. This call passes `(label, ent_name)`, so it produces `"person__albert einstein"` instead of `"albert_einstein__person"`; note the name also never gets its spaces normalized, because it lands in the type slot.

Effect: ingest a corpus, then `backfill_entity("Person")` — every backfilled entity gets an id in a different namespace from the extracted ones. The same real-world entity exists as two nodes and dedup / `MENTIONED_IN` linkage never joins them, i.e. it "silently disappears from deduplication" exactly as described in this PR's motivation.

galshubeli: **Medium — `backfill_entity` can still write reserved-label nodes** (about line 1101, just below the diff window).

`new_nodes.append(GraphNode(id=ent_id, label=label, properties=props))` uses the caller's `label` verbatim, with no `_reject_reserved_labels` call anywhere on this path.

Combined with the missing guard in `add_entity` (see my note on `graph_extraction.py:481`), this sequence writes real `:Document` nodes into the store:

```python
await rag.add_entity(Entity(label="Document"))
await rag.backfill_entity("Document")
```

That reproduces precisely the "11 documents reported as 106" corruption this PR sets out to prevent, through a path the fix doesn't cover.

graphrag_sdk/tests/test_graph_extraction.py
```diff
@@ -927,14 +992,49 @@ async def test_successful_extraction_reports_no_failures(self, ctx):
...
+        """The ontology path assigns entity types without going through
+        __init__, so it needs its own guard or the fix is bypassable."""
+        with pytest.raises(ValueError, match="reserved label"):
+            _reject_reserved_labels(["Person", "Document"])
 
-    def test_prompt_contains_short_token_filtering(self):
-        assert "1-2 characters" in VERIFY_EXTRACT_RELS_PROMPT
+    def test_store_and_extractor_share_one_definition(self):
+        """Two hardcoded copies would drift; the bug returns when they do."""
+        assert GraphStore._STRUCTURAL_LABELS is RESERVED_NODE_LABELS
```
galshubeli, line 1347: **Low — there's a third hardcoded copy of the structural labels, and this test doesn't cover it.**

The docstring says "two hardcoded copies would drift", but `retrieval/strategies/cypher_generation.py:39` has a third:

```python
_STRUCTURAL_LABELS = frozenset({"Chunk", "Document", "__Entity__"})
```

It's the one copy this assertion doesn't pin, so it's free to drift. Deriving it as `RESERVED_NODE_LABELS | {"__Entity__"}` would fold it into the same single definition and let this test cover all three.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: Fixed in d8eaf53a — verified.

`lowered in _ENTITY_STOPLIST and not (is_acronym and lowered in _PRONOUNS)` narrows the exemption to exactly the pronoun rows, which is the right shape. Moving `"one"` out of `_PRONOUNS` into the generic-noun block is a nice catch I had missed.

Ran the filter directly against the head commit:

- shell tokens `CD LS RM PS CAT ENV ETC VAR BIN USR TMP SSH SED AWK` — all rejected
- generic nouns `ONE MAN BOY NARRATOR AUTHOR` — all rejected
- real acronyms `AI US UK EU UN IT Go` — all still kept
- `_PRONOUNS <= _ENTITY_STOPLIST` holds, and `_PRONOUNS & _SHELL_TOKENS` is empty, so the two rule sets do not overlap

Residual, and it is the trade-off I asked for rather than a defect: uppercase pronouns `HE HER HIS ITS MY WE ME YOU` still pass, since `US` and `IT` are indistinguishable from the country and the language. Inherent to the approach; worth a line in the docstring if you want it on record.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: Fixed in d8eaf53a — verified, and the fix is better than what I proposed.

Rather than patching what discovery passes in, `entity_types=None` now means "no ontology, so no date rule", the extractors call `is_valid_entity_name(name)` with no types, and the ontology-dependent gate moved into `GraphExtraction._step1` where `entity_types` really is the ontology. That puts the rule where the knowledge lives instead of threading an ontology through NER.

Verified against head:

- `_parse_predictions(preds, _NER_ANCHOR_LABELS, ...)` returns all of `1984 747 1969 1776 1823`
- `LLMExtractor._parse_response` likewise
- `GraphExtraction` still drops `1823` when the ontology omits `Date` and keeps it when declared

The `_step1` filter placement checks out too: it sits after both the LLM-batch and local-extractor branches, `step2_indices` is rebuilt after it, empty inner lists were already a supported state (failed chunks append `[]`), and step 2 re-applies the gate at `graph_extraction.py:772,807` — so a dropped date cannot come back as a relationship endpoint.

galshubeli: Addressed in d8eaf53a in the code, but the PR body still needs the same edit.

The comment now says outright that the rule never fires on the default path and re-scopes the benchmark to "an ontology that omits `Date`" — that is exactly what I was asking for, and confirmed: `"Date" in DEFAULT_ENTITY_TYPES` is `True` and `is_valid_entity_name("1823", DEFAULT_ENTITY_TYPES)` is `True`.

The PR description §1 still reads as an out-of-the-box win, though, and that is what the next reviewer of the stack and the changelog will see. Worth carrying the same caveat up there.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
galshubeli: Fixed in d8eaf53a — verified.

Exact match now, so the store and the config check agree. Confirmed against head: `Document` and `Chunk` rejected; `document`, `cHuNk`, `DOCUMENT` allowed. `test_match_is_exact_like_the_store` pins the reasoning nicely. The `reserved` dict is gone in favour of a set as well.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
galshubeli: Fixed in d8eaf53a — verified, and the guard landed in the right place.

`reject_reserved_labels` now runs in `OntologyStore.register` before anything is persisted, and `_validate_proposal` appends an *error* rather than raising so the retry loop can ask the LLM for a different label — that is the better of the two options I mentioned.

I checked the `add_entity` hole specifically: `main.py:590` routes through `self._ontology_store.register(...)`, so it is covered by the same guard. `test_reserved_label_is_refused_before_anything_is_written` asserting no `MERGE` call is the right assertion.

Keeping the extractor-level check as a documented backstop for ontologies handed straight to `extract()` is the right call.

graphrag_sdk/src/graphrag_sdk/api/main.py
galshubeli: Fixed in d8eaf53a — verified.

`ontology_labels = [e.label for e in self._global_ontology.entities]`, hoisted out of the per-entity loop. `test_date_gate_uses_the_ontology_not_the_target_label` covers the case I described: backfilling `Product` in a graph declaring `Date` keeps "747" with `values_skipped == 0`.

graphrag_sdk/src/graphrag_sdk/api/main.py
galshubeli: Fixed in d8eaf53a — verified.

`compute_entity_id(ent_name, label)` now matches every other call site. Confirmed against head: `compute_entity_id("Albert Einstein", "Person")` returns `albert_einstein__person`, so the name gets its spaces normalised and backfilled entities share a namespace with extracted ones. The test asserting `compute_entity_id("Bob", "Person") == "bob__person"` and `rel.start_node_id == node.id` pins both halves.

galshubeli: Fixed in d8eaf53a — verified.

`reject_reserved_labels([label])` at the top of `backfill_entity`, before any store call. With the `register`-time guard the label should never reach the ontology in the first place, but keeping this as defence-in-depth for graphs written before the guard existed is the right instinct, and the comment says so.

`test_reserved_label_is_refused` asserting `upsert_nodes.assert_not_awaited()` is the assertion that matters — it proves nothing is written, not merely that it raises.

graphrag_sdk/tests/test_graph_extraction.py
galshubeli: Fixed in d8eaf53a — verified.

`_STRUCTURAL_LABELS = RESERVED_NODE_LABELS | {"__Entity__"}` in the Cypher generator, and the test now pins all three definitions instead of two. Grepped the head commit for remaining `{"Chunk", "Document"}` literals: one hit, the `RESERVED_NODE_LABELS` definition itself. Single source of truth.

graphrag_sdk/.venv
```diff
@@ -0,0 +1 @@
+/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
```
galshubeli: **Blocker — a symlink to your local machine got committed here.**

This file is mode `120000` (a symlink), pointing at `/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv`. Two problems:

- It checks out **dangling** on every other machine — confirmed in a fresh worktree on Linux, where `graphrag_sdk/.venv` becomes a broken link.
- On a macOS checkout that happens to have that path, it resolves silently to *your* virtualenv, which is worse than a broken link because nothing complains.

Nothing in the repo catches it: the root `.gitignore` has `.venv/`, which matches a directory, not a symlink named `.venv`. Since CI doesn't run on this PR (see below), it would ride the stack all the way to `main`.

```
git rm --cached graphrag_sdk/.venv
```

Worth also relaxing the ignore rule to `.venv` (no trailing slash) so a symlink can't slip through the same way again.

---

Two other things to fold in before merge, neither of them code problems:

**The PR description now contradicts the code in two places.** §2 still says an all-caps token "skips the stop-word check" — after d8eaf53a it skips only the pronoun rows, which is the whole point of that fix. §4 still says reserved labels are "rejected case-insensitively at configuration time" — the match is now exact, and the primary guard moved to `OntologyStore.register`, before persistence. §1 also still reads as an out-of-the-box precision win when `DEFAULT_ENTITY_TYPES` declares `Date` (the code comment now says this correctly, the body doesn't). Worth fixing since that description is what the next reviewer of the stack, and the changelog, will read.

**Heads-up that CI has never run on this PR.** `.github/workflows/ci.yml` triggers on `pull_request: branches: [main, staging]`, and this targets `feat/extraction-gliner-perf`, so `d8eaf53a` has exactly one check run (the Copilot reviewer) — no lint, no tests, on any Python version. I ran the suite locally at this commit: **1196 passed, 42 skipped**, and `ruff format --check src/` is clean. But nothing has verified it on 3.10 or 3.12, so whoever merges the stack should keep that in mind.

graphrag_sdk/tests/test_backfill.py
```diff
@@ -389,6 +389,46 @@ async def test_creates_entity_and_mention_edge(self, rag):
...
+
+    @pytest.mark.asyncio
+    async def test_date_gate_uses_the_ontology_not_the_target_label(self, rag):
+        """Backfilling `Product` in a graph whose ontology declares `Date` must
+        keep numeric names ("747") exactly as normal extraction would; only the
+        ontology as a whole says whether date-shaped names are wanted."""
+        rag._global_ontology = Ontology(
+            entities=[Entity(label="Product"), Entity(label="Date")],
+        )
+        rag.llm = MockLLM(
```
galshubeli: **Nit — `ruff format` wants two of the new test blocks reformatted.**

Two spots, both new in d8eaf53a:

- here: this `MockLLM(...)` call fits on one line at the configured width, so ruff collapses the three lines into one
- `tests/test_ontology_store.py:170`: no blank line between the end of `test_each_attribute_becomes_a_property_node` and the new `@pytest.mark.asyncio`

One `ruff format tests/` run fixes both. The same files pass the check at `9222a143`, and all seven changed `src/` files are clean.

To be clear about the severity, since I nearly filed this as a gate: **CI will not catch it and will not fail on it** — `ci.yml` runs `ruff check src/` and `ruff format --check src/` only, never `tests/`. So it's cosmetic, and only worth doing because the commit right before this one in the series is `style: ruff format`.

galshubeli (approved):

PR #320, opened 2026-09-07
Extraction: GLiNER windows over long chunks, per-model thresholds, 'Unknown' band

Naseem77:
## Extraction — GLiNER reads the whole chunk, per-model thresholds, "Unknown" band

Part of FalkorDB/research#88. Stacked on #319.

### 1. GLiNER reads the whole chunk
**Problem:** GLiNER truncates input at `config.max_len` silently — for `gliner_medium-v2.1` a probe entity at word 388 is always found, at word 389 never. Everything past that in a chunk was invisible; disabling the fix on the benchmark drops ceiling recall 0.805 → 0.621.
**Fix:** text longer than the window is processed as overlapping windows (overlap 48 word-tokens, above the model's 12-word `max_width`) and merged. Short text takes the old fast path.

### 2. Per-model confidence thresholds
**Problem:** one hard-coded 0.75 for every model. Scores are not comparable across models — the bi-encoders return 2 entities for a whole corpus at 0.75.
**Fix:** `DEFAULT_THRESHOLDS` (medium / large / medium-v2.5: 0.75; bi-small / bi-base: 0.50); `threshold=None` looks the model up, unknown models fall back to 0.5 with a warning. Default model stays `urchade/gliner_medium-v2.1` (bi-small measured better in August but returns 0 entities under gliner 0.2.27 / transformers 5.6).

### 3. "Unknown" band 25 % below the threshold
**Problem:** everything under the cutoff was discarded inside GLiNER; borderline real names were lost.
**Fix:** `candidate_threshold` defaults to `threshold × 0.75` (0.5625–0.75 for the default model). Spans in that band reach the step-2 LLM labelled `Unknown`, which re-types or drops them against the text; below the band is still discarded. `candidate_threshold=None` turns it off.
**Measured** (same code, band off vs on): NER spans 1,828 → 2,468; 0 `Unknown` nodes reached the graph (the LLM typed every kept one); entity recall 0.535 → 0.571 at precision 0.605 → 0.572 (F1 0.568 → 0.572); triple relaxed F1 0.228 → 0.236; answer accuracy 30.4 → 32.4 %; +8 % ingest cost. A wide fixed floor of 0.30 tested earlier was net negative (F1 −0.027), which is why the band is relative and narrow.
*Measured on `2e133d4`, which merged cross-window spans with the original label-scoped rule. Review rounds 1–2 (`40993b2`, `133a079`) replaced that merge — containment now loses to the container regardless of label, partial overlaps resolve by score — which changes which spans reach step 2, so these figures have not been re-measured against the current merge.*

### Verification
Full suite: **1203 passed, 41 skipped**; `ruff check` / `ruff format --check` clean on `src/`, `tests/test_entity_extractors.py` ruff-check clean.

## Summary by CodeRabbit

- **New Features**
  - GLiNER extraction now supports configurable confidence thresholds, candidate thresholds, and model-specific defaults.
  - Long text is processed in overlapping windows while preserving and merging entity spans.
  - Low-confidence candidates may be labeled “Unknown” for follow-up processing or discarded based on configuration.

- **Documentation**
  - Updated extractor and strategy references to describe the new parameters, validation rules, windowed processing, and confidence behavior.

### DIFF
--- a/docs/extraction.mdx
+++ b/docs/extraction.mdx
@@ -81,7 +81,8 @@ A local transformer model that runs on your machine — no API calls needed.
 from graphrag_sdk import GLiNERExtractor
 
 extractor = GLiNERExtractor(
-    threshold=0.75,                          # confidence threshold (below = "Unknown")
+    threshold=0.75,                          # at or above: typed entity
+    candidate_threshold=0.5625,              # band below threshold kept as "Unknown" (default: 25 % below threshold)
     model_name="urchade/gliner_medium-v2.1", # HuggingFace model
 )
 ```
@@ -90,7 +91,7 @@ extractor = GLiNERExtractor(
 1. The model is loaded lazily on first use (thread-safe via `threading.Lock`)
 2. Runs inference via `asyncio.to_thread()` to avoid blocking the event loop
 3. Returns predictions with character-level spans — more precise than LLM spans
-4. Entities below the confidence threshold are labeled `"Unknown"` (not discarded)
+4. Entities in the band just below the confidence threshold (default: 25 % below it, e.g. 0.5625–0.75) are labeled `"Unknown"` and handed to the step-2 LLM, which re-types or drops them against the text; anything below the band is discarded. Pass `candidate_threshold=None` to disable the band.
 
 **Best for:** Production pipelines where you want fast, cheap NER without API costs.
 
@@ -198,7 +199,7 @@ Person, Organization, Technology, Product, Location,
 Date, Event, Concept, Law, Dataset, Method
 ```
 
-Entities that don't match any type (or fall below the confidence threshold) are labeled `"Unknown"`.
+Entities that don't match any type (or score in the band just below the confidence threshold) are labeled `"Unknown"`.
 
 ### Customizing the Ontology
 
--- a/docs/extraction.mdx
+++ b/docs/extraction.mdx
@@ -81,17 +81,22 @@ A local transformer model that runs on your machine — no API calls needed.
 from graphrag_sdk import GLiNERExtractor
 
 extractor = GLiNERExtractor(
-    threshold=0.75,                          # at or above: typed entity
+    threshold=0.75,                          # at or above: typed entity (default: None = per-model value)
     candidate_threshold=0.5625,              # band below threshold kept as "Unknown" (default: 25 % below threshold)
-    model_name="urchade/gliner_medium-v2.1", # HuggingFace model
+    model_name="urchade/gliner_medium-v2.1", # HuggingFace model (default)
+    window_tokens=None,                      # words per inference window (default: derived from the model's max_len)
+    window_overlap=48,                       # words shared by consecutive windows
 )
 ```
 
 **How it works:**
 1. The model is loaded lazily on first use (thread-safe via `threading.Lock`)
 2. Runs inference via `asyncio.to_thread()` to avoid blocking the event loop
-3. Returns predictions with character-level spans — more precise than LLM spans
-4. Entities in the band just below the confidence threshold (default: 25 % below it, e.g. 0.5625–0.75) are labeled `"Unknown"` and handed to the step-2 LLM, which re-types or drops them against the text; anything below the band is discarded. Pass `candidate_threshold=None` to disable the band.
+3. Text longer than the model's input limit (`config.max_len`, 384 words for the default model) is processed as overlapping windows and merged back into non-overlapping spans, so the tail of a long chunk is not truncated. Short text takes a single-pass fast path.
+4. Returns predictions with character-level spans — more precise than LLM spans
+5. Entities in the band just below the confidence threshold (default: 25 % below it, e.g. 0.5625–0.75) are labeled `"Unknown"` and handed to the step-2 LLM, which re-types or drops them against the text; anything below the band is discarded. Pass `candidate_threshold=None` to disable the band.
+
+Thresholds are model-specific: `threshold=None` (the default) selects the measured value for `model_name` (`0.75` for the `urchade/gliner_*` models, `0.5` for the `knowledgator/gliner-bi-*` bi-encoders); unknown models fall back to `0.5` with a warning. Invalid configurations raise `ValueError` at construction: `threshold` outside `[0, 1]`, `candidate_threshold` outside `[0, threshold]`, `window_tokens <= 0`, or `window_overlap` negative or not smaller than the window.
 
 **Best for:** Production pipelines where you want fast, cheap NER without API costs.
 
--- a/docs/extraction.mdx
+++ b/docs/extraction.mdx
@@ -96,7 +96,7 @@ extractor = GLiNERExtractor(
 4. Returns predictions with character-level spans — more precise than LLM spans
 5. Entities in the band just below the confidence threshold (default: 25 % below it, e.g. 0.5625–0.75) are labeled `"Unknown"` and handed to the step-2 LLM, which re-types or drops them against the text; anything below the band is discarded. Pass `candidate_threshold=None` to disable the band.
 
-Thresholds are model-specific: `threshold=None` (the default) selects the measured value for `model_name` (`0.75` for the `urchade/gliner_*` models, `0.5` for the `knowledgator/gliner-bi-*` bi-encoders); unknown models fall back to `0.5` with a warning. Invalid configurations raise `ValueError` at construction: `threshold` outside `[0, 1]`, `candidate_threshold` outside `[0, threshold]`, `window_tokens <= 0`, or `window_overlap` negative or not smaller than the window.
+Thresholds are model-specific: `threshold=None` (the default) selects the measured value for `model_name` (`0.75` for the `urchade/gliner_*` models, `0.5` for the `knowledgator/gliner-bi-*` bi-encoders); unknown models fall back to `0.5` with a warning. Invalid configurations raise `ValueError`: `threshold` outside `[0, 1]`, `candidate_threshold` outside `[0, threshold]`, `window_tokens` not a positive integer, or `window_overlap` not a non-negative integer — all at construction. `window_overlap` must also be smaller than the window; that is checked at construction when `window_tokens` is given explicitly, otherwise on the first inference call, once the window has been derived from the model.
 
 **Best for:** Production pipelines where you want fast, cheap NER without API costs.
 
--- a/graphrag_sdk/.venv
+++ b/graphrag_sdk/.venv
@@ -1,0 +1,1 @@
+/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
--- a/graphrag_sdk/.venv
+++ b/graphrag_sdk/.venv
@@ -1,1 +1,0 @@
-/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
--- a/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
+++ b/graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
@@ -7,7 +7,6 @@
 from __future__ import annotations
 
 import asyncio
-import bisect
 import json
 import logging
 import re
@@ -41,6 +40,12 @@ UNKNOWN_LABEL = "Unknown"
 
 #: Sentinel for "use the class default" where ``None`` is itself a valid value.
 _DEFAULT: Any = object()
+
+
+def _is_int(value: Any) -> bool:
+    """``bool`` is an ``int`` subclass; window sizes must be real integers."""
+    return isinstance(value, int) and not isinstance(value, bool)
+
 
 MIN_NAME_LEN = 2  # single-char names are noise
 MAX_NAME_LEN = 80  # descriptions masquerading as names
@@ -475,16 +480,18 @@ class GLiNERExtractor(EntityExtractor):
     corpus**, with no error raised.
 
     Args:
-        threshold: Confidence threshold (0-1). Below this → "Unknown".
+        threshold: Confidence threshold (0-1). Spans scoring at or above it
+            are typed; spans between ``candidate_threshold`` and it are
+            labelled ``"Unknown"``; anything lower is discarded by the model.
             ``None`` (default) selects the value measured for ``model_name``.
         model_name: HuggingFace model name for GLiNER.
         window_tokens: Word-tokens per inference window. ``None`` derives it
             from the model's own ``config.max_len`` minus a safety margin.
             An explicit value must be a positive integer.
-        window_overlap: Word-tokens shared between consecutive windows. Must
-            satisfy ``0 <= window_overlap < window_tokens`` (checked against
-            the derived window on first use when ``window_tokens`` is
-            ``None``) and should exceed the model's ``max_width`` (longest
+        window_overlap: Word-tokens shared between consecutive windows. An
+            integer satisfying ``0 <= window_overlap < window_tokens`` (checked
+            against the derived window on first use when ``window_tokens`` is
+            ``None``); it should exceed the model's ``max_width`` (longest
             representable entity, 12 words by default) or entities on a
             boundary are lost.
         candidate_threshold: Floor of the ``"Unknown"`` band. Default: 25 %
@@ -495,7 +502,8 @@ class GLiNERExtractor(EntityExtractor):
     Raises:
         ValueError: If ``threshold`` is outside ``[0, 1]``, if
             ``candidate_threshold`` is outside ``[0, threshold]``, or if the
-            window parameters are degenerate (``window_tokens <= 0``,
+            window parameters are degenerate (``window_tokens`` or
+            ``window_overlap`` not an ``int``, ``window_tokens <= 0``,
             ``window_overlap < 0`` or ``window_overlap >= window_tokens``).
     """
 
@@ -574,15 +582,16 @@ class GLiNERExtractor(EntityExtractor):
         # ``_CACHE_LOCK`` and inference deliberately takes no lock at all.
         # See ``_predict_sync`` for the thread-safety evidence.
         self._lock = threading.Lock()
-        if window_tokens is not None and (isinstance(window_tokens, bool) or window_tokens <= 0):
+        if window_tokens is not None and not (_is_int(window_tokens) and window_tokens > 0):
             # ``words[begin:begin + 0]`` is empty on the first iteration, so a
-            # non-positive window would return zero entities with no error.
-            raise ValueError(f"window_tokens ({window_tokens}) must be a positive integer")
-        if window_overlap < 0:
+            # non-positive window would return zero entities with no error; a
+            # non-int only fails later, inside ``range()`` at inference time.
+            raise ValueError(f"window_tokens ({window_tokens!r}) must be a positive integer")
+        if not (_is_int(window_overlap) and window_overlap >= 0):
             # A negative overlap makes ``step > window`` and leaves word ranges
             # that never reach the model — the silent truncation this class
             # exists to prevent, reintroduced by configuration.
-            raise ValueError(f"window_overlap ({window_overlap}) must be >= 0")
+            raise ValueError(f"window_overlap ({window_overlap!r}) must be a non-negative integer")
         if window_tokens is not None:
             self._check_overlap(window_overlap, window_tokens)
         self._window_tokens = window_tokens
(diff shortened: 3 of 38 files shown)


galshubeli (commented):
Read the diff plus the whole post-change `entity_extractors.py`, the downstream `Unknown` handling, and the installed gliner 0.2.25 source. The premises hold up: `max_len` truncation is at the *word* level (`data_processing/processor.py:490`), so a word-based window is the right unit, and predicted `start`/`end` are exclusive char offsets, so the remapping is correct. Exercised against a stub model, a 1000-word text gave 4 windows and recovered 11/11 predictions identically to the unwindowed run, zero offset mismatches. Comments inline — the first two are the ones I'd want fixed before merge.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -427,25 +430,137 @@ class GLiNERExtractor(EntityExtractor):
...
+    #: per-model threshold instead of assuming one score scale.
+    CANDIDATE_BAND = 0.25
+
     def __init__(
         self,
-        threshold: float = 0.75,
-        model_name: str = "urchade/gliner_medium-v2.1",
+        threshold: float | None = None,
+        model_name: str | None = None,
+        window_tokens: int | None = None,
```
galshubeli, line 582: **Window params are never validated — three silent failure modes.**

`__init__` validates `candidate_threshold` but nothing about `window_tokens`/`window_overlap`. Measured against a stub model on a 2000-word text:

- `window_tokens=30` with the default `window_overlap=48` → `step = max(1, 30-48) = 1` → **982 sequential `predict_entities` calls** instead of ~7. Under real inference that's an effective hang.
- `window_tokens=0` (e.g. computed from a config) → `words[begin:begin+0]` is empty on the first iteration, `break` fires, and the extractor returns **zero entities with no error**.
- a negative `window_overlap` makes `step > window`, leaving uncovered word ranges that are never sent to the model — the exact silent-truncation bug this PR fixes, reintroduced by configuration.

Suggest rejecting `window_tokens <= 0` and requiring `0 <= window_overlap < window`.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -489,9 +604,66 @@ def _get_shared_model(cls, model_name: str) -> Any:
...
+        best: dict[tuple[int, int, str], dict[str, Any]] = {}
+        for p in preds:
+            key = (p["start"], p["end"], p["label"])
+            prev = best.get(key)
+            if prev is None or p.get("score", 0.0) > prev.get("score", 0.0):
+                best[key] = p
+
+        kept: list[dict[str, Any]] = []
+        for p in best.values():
+            contained = any(
```
galshubeli: **`_merge`'s containment rule is label-scoped, so window-edge fragments leak through.**

The `contained` predicate requires `q["label"] == p["label"]`, and it only drops *nested* spans, never partially overlapping ones. Verified with a stub: "Acme Corporation Ltd" straddles window A's right edge, A returns the truncated `"Acme"` (label `product`, 0.6), B returns the full `"Acme Corporation Ltd"` (label `organization`, 0.9). Labels differ → `contained` is False → the merge keeps **both**. A pure truncation artifact becomes an extra entity, and at 0.6 it lands in the new Unknown band, so it also reaches the step-2 LLM as a candidate.

Worth noting this is a new shape for downstream code: `flat_ner=True` guaranteed non-overlapping spans in the unwindowed path, and the windowed path now emits overlapping/nested ones. Resolving by character overlap (prefer the longer / higher-scoring span) rather than by `(label, containment)` would close it.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -489,9 +604,66 @@ def _get_shared_model(cls, model_name: str) -> Any:
                 cls._MODEL_CACHE[model_name] = cached
         return cached
 
+    def _resolve_window(self, model: Any) -> int:
+        """Window size in word-tokens, derived from the model if not set."""
+        if self._window_tokens is not None:
```
galshubeli, line 687: Same root as the validation comment above: the derived branch clamps to at least 64 word-tokens, but the explicit branch returns the caller's value verbatim. That asymmetry is what makes `window_tokens=30`/`window_tokens=0` reachable, and it means the two branches disagree about what counts as a valid window. Applying the same floor (or the `__init__` validation) in both would be enough.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -427,25 +430,137 @@ class GLiNERExtractor(EntityExtractor):
...
+                    "%.2f. Thresholds are not comparable between models, so "
+                    "tune this before relying on the results.",
+                    self._model_name,
+                    self._FALLBACK_THRESHOLD,
+                )
         self._threshold = threshold
-        self._model_name = model_name
+        if candidate_threshold is _DEFAULT:
+            candidate_threshold = round(threshold * (1.0 - self.CANDIDATE_BAND), 4)
+        if candidate_threshold is not None and candidate_threshold > threshold:
```
galshubeli: `candidate_threshold` is only bounded from above. `GLiNERExtractor(candidate_threshold=-1)` is accepted (verified: `_candidate_threshold == -1`), so the model is queried with `threshold=-1` and returns essentially every candidate span in the chunk — all of which pass as `"Unknown"`, which `ingestion/pipeline.py:408` explicitly whitelists, so the junk reaches step 2 and then the graph. `threshold=0` does the same via a `0.0` floor. A `0 <= candidate_threshold <= threshold <= 1` check would cover both.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -512,17 +684,50 @@ def _predict_sync(self, text: str, entity_types: list[str]) -> list[dict[str, An
...
+                p = dict(p)
+                p["start"] += lo
+                p["end"] += lo
+                p["text"] = text[p["start"] : p["end"]]
+                out.append(p)
+            if begin + window >= len(words):
+                break
+
+        merged = self._merge(out)
+        logger.debug(
```
galshubeli, line 832: The window count here is `(len(words) - 1) // step + 1` = `ceil(len/step)`, but the loop breaks at the first `begin` where `begin + window >= len(words)`, i.e. `ceil((len - window)/step) + 1` windows. With `window_tokens=350, window_overlap=300` (step 50) on a 400-word text the loop runs 2 windows and this logs 8.

Harmless at the defaults (step 302 ≈ window), but it misleads exactly the person who has raised the overlap and is reading these logs to see the effect.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -427,25 +430,137 @@ class GLiNERExtractor(EntityExtractor):
     so a single instance can be safely shared across concurrent
     ``asyncio.to_thread`` calls (e.g. parallel doc ingestion).
 
+    GLiNER has a hard input limit (``config.max_len``) and truncates anything
+    beyond it **silently** — no exception, no warning, the tail simply never
+    reaches the model. Measured on ``gliner_medium-v2.1`` (``max_len`` 384): a
+    probe entity placed at word-token 388 is always returned, at 389 never.
+    The default model's limit is 2048, but real documents still exceed it —
```
galshubeli: Two small doc corrections in this docstring:

- This line says the default model's limit is 2048, but line 435 says "Measured on `gliner_medium-v2.1` (`max_len` 384)" and the `DEFAULT_MODEL` comment (line 496) puts 2048 on `gliner-bi-small-v2.0` and 384 on the medium model — and `DEFAULT_MODEL` *is* `gliner_medium-v2.1`. Since `_WINDOW_MARGIN = 34` is justified relative to that number, the wrong figure will misdirect whoever next touches the window math.
- "truncates … **silently** — no exception, no warning" isn't true of the installed gliner: `data_processing/processor.py:491` emits `UserWarning: Sentence of length N has been truncated to 384`. The truncation is still the bug; it just isn't silent.

graphrag_sdk/tests/test_entity_extractors.py
```diff
@@ -223,3 +223,44 @@ def test_inference_takes_no_lock(self):
             line for line in src.splitlines() if not line.strip().startswith("#")
         )
         assert "self._lock" not in code
+
+
+class TestGLiNERCandidateBand:
```
galshubeli, line 319: `TestGLiNERCandidateBand` covers the threshold/band arithmetic, but `_predict_body`, `_merge` and `_resolve_window` — ~55 new lines holding all of the offset remapping and dedup logic — have no coverage. Both of the issues I flagged above live in that untested code.

A stub model exposing `data_processor.words_splitter` and a regex `predict_entities` was enough for me to pin all of this locally: offsets round-trip (`text[start:end]` equals the reported entity text), windowed output equals unwindowed output on a long text, boundary entities aren't duplicated, and the degenerate window configs raise instead of hanging or returning nothing. Happy to hand over the stub if useful.

docs/extraction.mdx
```diff
@@ -90,7 +91,7 @@ extractor = GLiNERExtractor(
 1. The model is loaded lazily on first use (thread-safe via `threading.Lock`)
 2. Runs inference via `asyncio.to_thread()` to avoid blocking the event loop
 3. Returns predictions with character-level spans — more precise than LLM spans
-4. Entities below the confidence threshold are labeled `"Unknown"` (not discarded)
+4. Entities in the band just below the confidence threshold (default: 25 % below it, e.g. 0.5625–0.75) are labeled `"Unknown"` and handed to the step-2 LLM, which re-types or drops them against the text; anything below the band is discarded. Pass `candidate_threshold=None` to disable the band.
```
galshubeli: This page documents the band, but two other pages are now stale or misleading:

- `docs/api-reference.mdx:603` still reads `GLiNERExtractor(threshold=0.75, model_name="urchade/gliner_medium-v2.1")`. Both defaults are now `None`, and `window_tokens`, `window_overlap` and the behaviour-changing `candidate_threshold` default aren't mentioned.
- `docs/strategies.mdx:305` (`GLiNERExtractor(threshold=0.6)`) is now silently a band-on configuration with a 0.45 floor, so that documented example extracts a different set of entities than it did before this PR, with no prose saying so.

graphrag_sdk/.venv
```diff
@@ -0,0 +1 @@
+/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv
```
galshubeli: **Blocking — this got committed by accident.** It's not a directory but a symlink (mode `120000`) pointing at `/Users/naseemali/Documents/GitHub/GraphRAG-SDK/graphrag_sdk/.venv`.

Anyone checking out the branch gets a dangling symlink where tooling expects a virtualenv, and your local filesystem layout ends up in the repo history. It slipped past `.gitignore` because line 58 is `.venv/` with a trailing slash, which matches a directory but not a symlink of that name.

```
git rm --cached graphrag_sdk/.venv
```

and a slash-less `.venv` ignore rule so it can't recur.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -489,9 +641,92 @@ def _get_shared_model(cls, model_name: str) -> Any:
...
+        back with a different label or a higher score than the whole entity
+        seen by the neighbouring window. Exact duplicates collapse to the
+        highest-scoring copy.
+
+        O(n log n): kept spans are held in a sorted list and each candidate
+        checks its neighbours by bisection.
+        """
+        ranked = sorted(
+            preds,
+            key=lambda p: (p["end"] - p["start"], p.get("score", 0.0)),
```
galshubeli: **Blocking — the rewrite fixed the label bug but introduced a coverage bug.** Considering candidates longest-first and dropping *any* overlapper is greedy, not maximum-coverage, and the score never gets a say. Measured against a stub:

```
in : ("Bank of America",         0-15, 0.88)
     ("America Online Services", 8-31, 0.71)   <- longest, lowest score
     ("Online Services Inc",    16-35, 0.85)
out: ("America Online Services", 8-31)          -> 1 kept, of 2 possible
```

The least confident span wins on length alone and takes two disjoint higher-confidence entities with it. The old label-scoped code kept all three, so this is a regression against pre-PR behaviour — and it's reachable exactly where the new windowing lives: `flat_ner` prevents overlap *within* a window, so these come from two windows reading the shared 48-word overlap region differently.

I'd hold on this one mainly because of what the PR is for: the whole justification is "we were silently losing entities", and this silently loses entities in the path the PR adds.

The truncation case you're actually targeting is *containment* — "Acme" nested inside "Acme Corporation Ltd". So: let length win only when one span contains the other, and fall back to score for partial overlaps. That keeps the Acme fix (verified — that case now correctly yields only `("Acme Corporation Ltd", organization)`) without discarding disjoint neighbours.

Separately, on the same key: exact ties on `(length, score)` currently resolve by window emission order, since `sorted` is stable. Over 200 shuffles of one 7-span input I got two distinct outputs (`{(40,50),(50,60)}` vs `{(45,55)}`). It's deterministic for fixed text and model, so not flaky in production, but which entity survives is an artifact of where the window boundary fell — adding `-p["start"]` to the key makes it explicit.

graphrag_sdk/tests/test_entity_extractors.py
```diff
@@ -223,3 +225,283 @@ def test_inference_takes_no_lock(self):
...
+
+    def test_merge_same_span_two_labels_keeps_best_score(self):
+        merged = GLiNERExtractor._merge([
+            {"text": "Paris", "label": "person", "score": 0.6, "start": 0, "end": 5},
+            {"text": "Paris", "label": "location", "score": 0.9, "start": 0, "end": 5},
+            {"text": "Paris", "label": "location", "score": 0.8, "start": 0, "end": 5},
+        ])
+        assert [(p["label"], p["score"]) for p in merged] == [("location", 0.9)]
+
+    def test_merge_scales(self):
```
galshubeli, line 624: Non-blocking, but this test passes for the wrong reason, and the `O(n log n)` claim at `entity_extractors.py:692` isn't right either.

`kept_starts.insert(i, …)` is an O(n) memmove — `bisect` only makes the *search* logarithmic, so the merge is O(n²) worst case. This test generates spans in **ascending** start order with identical lengths and scores, so the stable sort preserves that order and every insert lands at the end, which is the one input shape that really is linear. Measured at n = 5k/10k/20k:

```
ascending (this test):  5.4 / 11.3 /  23.1 ms   (2x per doubling)
descending:            12.1 / 35.1 / 116.2 ms   (~3x)
shuffled:              12.9 / 40.2 / 106.2 ms   (~3x)
```

Practically irrelevant at real per-chunk n (hundreds of spans, sub-ms) and the 2.0 s bound has 87x margin so it won't flake — but the test doesn't guard what it claims. Either shuffle the input, or reword the docstring to O(n²) worst case / O(n log n) typical.

graphrag_sdk/tests/test_entity_extractors.py
```diff
@@ -3,6 +3,8 @@
 from __future__ import annotations
```
galshubeli, line 3: Non-blocking: this file isn't ruff-clean under the repo's own config (line-length 100, `select = ["E","F","I","W","UP"]`), and `ruff format --check` would reformat it:

```
tests/test_entity_extractors.py:3:1:     I001 Import block is un-sorted or un-formatted
tests/test_entity_extractors.py:406:101: E501 Line too long (102 > 100)
tests/test_entity_extractors.py:453:101: E501 Line too long (117 > 100)
```

`entity_extractors.py` itself is clean and already formatted — just a missed pass, but the PR description does claim ruff clean.

galshubeli (commented):
Re-reviewed `40993b2` against the new code rather than the commit message, and ran the tests: `tests/test_entity_extractors.py` **48 passed, 1 skipped**; full suite **1196 passed, 41 skipped**, no regressions. Seven of the eight findings are genuinely fixed — per-finding verdicts are in the threads. The splitter removal also checks out: `_word_spans` is called once per chunk rather than per window, the fallback costs ~3.7 µs, and dropping the instance mutation strictly improves the thread-safety story.

Two blockers below (the `.venv` symlink and a regression the `_merge` rewrite introduced), then three non-blocking notes.

One thing worth flagging separately: the benchmark figures in the PR description (recall 0.535 → 0.571, F1 0.568 → 0.572, answer accuracy 30.4 → 32.4 %) were measured on `2e133d4`, with the old label-scoped merge. The new merge changes which spans survive into step 2, so those numbers no longer describe the code up for approval. Not asking for a full re-run to merge — but once the merge fix below is in, either a spot-check or a note in the description, so the measurements match the code.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: **Verified fixed.** `__init__` now rejects `window_tokens <= 0` (including `bool`), negative `window_overlap`, and `overlap >= window`. Probed each case: `window_tokens=30` against the default overlap 48 raises instead of collapsing `step` to 1; `0`, `-5`, `True`, `False` all rejected; `window_overlap=-10` rejected. Ran a coverage probe over 2000 words across four configs — 0 uncovered word indices in all of them, so the silent-gap case is closed too.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: **Verified fixed** — `0 <= candidate_threshold <= threshold` is enforced, and you added a `0 <= threshold <= 1` check on top, which closes the `threshold=0` variant as well.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: **Verified fixed.** Replacing the formula with a real per-call counter is the right call — it can no longer drift from the loop. Nice that `test_window_count_in_log_matches_calls` asserts both `len(stub.calls) == 2` and the log string at overlap 300 / window 350, which is exactly the config that was wrong before.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: **Verified fixed.** Now reads "`gliner_medium-v2.1` (the default model, `max_len` 384)" and "The bi-encoder models allow 2048", consistent with the `DEFAULT_MODEL` comment at line 508 — so `_WINDOW_MARGIN` is justified against the right number. The "silently" wording is corrected to name the `UserWarning` too.

docs/extraction.mdx
galshubeli: **Verified fixed** across all three pages: `api-reference.mdx:603` shows the real signature, `strategies.mdx:302` now explains the 0.45–0.6 band on the `threshold=0.6` example, and this page documents the windowing and the `ValueError` rules.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: The label-scoping is fixed — verified the exact case from this thread: "Acme Corporation Ltd" straddling a window edge now yields only `("Acme Corporation Ltd", organization)`, and abutting spans `[0,4)+[4,9)` both survive, so no off-by-one at the boundary.

But the rewrite introduced a different way to lose entities — a long low-confidence span now deletes disjoint higher-confidence ones. Details and a repro on the sort key: https://github.com/FalkorDB/GraphRAG-SDK/pull/320#discussion_r3967435691 — that one I would fix before merging.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
galshubeli: **Partially fixed, and I would accept it as-is.** The reachability is closed — the hang and silent-empty cases now raise — and `_resolve_window` validates both branches, which was the real problem.

The floor itself is still explicit-path-only, so `window_tokens=1, window_overlap=0` is accepted and produces 2000 `predict_entities` calls on a 2000-word text (measured). But that is now a self-consistent, explicitly-chosen value rather than a surprise interaction with a default, and the docstring says "must be a positive integer". Not worth blocking on.

graphrag_sdk/tests/test_entity_extractors.py
galshubeli: **Fixed** — 24 tests, and they assert on real behaviour rather than on the stub: offsets round-trip (`text[start:end] == p["text"]`), windowed output equals unwindowed on 2000 words with an exact call count of 7, boundary dedupe, five parametrized degenerate configs, and `vars(ex)` unchanged after inference. This is the coverage the change needed; the stub design makes the windowing properties checkable without loading a model, which is the right trade.

Two follow-ups on the new tests, both non-blocking, commented inline: `test_merge_scales` passes for the wrong reason (its input is pre-sorted, which is the one linear shape), and the file has 3 ruff errors.

galshubeli (approved):

PR #321, opened 2026-09-07
Extraction: default relation vocabulary and no early stop in relation extraction

Naseem77:
## Extraction — a default relation vocabulary, and no more early stop

Part of FalkorDB/research#88. Stacked on #320. Both changes are to the one step-2 prompt (entities + relations per chunk).

### 1. `DEFAULT_RELATION_TYPES`
**Problem:** entities always had a default ontology; relations had none — the prompt said "invent a descriptive label", so the same relation got a different name in every chunk. 447 distinct relation labels on the benchmark, only 17 % matching the answer key's 30; exact triple F1 0.065.
**Fix:** a 31-label default vocabulary offered to the LLM, the exact counterpart of `entity_types` — guidance, not a filter (labels outside the list are kept; `relation_types=[]` restores open vocabulary; a declared ontology overrides it). Vocabulary 447 → ~94, exact triple F1 2×; better on 5 of 5 unseen domains.

### 2. The relation extractor no longer stops early
**Problem:** the model treats the task as a summary — ~12 relations per chunk, then stops, using 2.5k of a 16k reply budget. Giving it twice the text grew the reply 1.3 %.
**Fix:** three prompt sentences ("exhaustive extraction, not a summary; there is no maximum; never end the list early"). +15 % relations for +4 % ingest time, and the best exact score of every variant tried — a second "what did you miss?" call gave more relations at 4.7× the time and a worse score, and is not in this PR.

### Verification
Full suite: **1179 passed, 41 skipped**; ruff clean.

## Summary by CodeRabbit

- **New Features**
  - Graph extraction now supports configurable relationship vocabularies through `relation_types`.
  - A default set of standardized relationship labels improves consistency while allowing open-vocabulary extraction when explicitly selected.
  - Declared ontology relationships remain enforced, and relationship extraction now more thoroughly identifies connections using verified entities.

- **Documentation**
  - Updated API and extraction guides to explain relationship vocabulary options, defaults, and ontology behavior.

- **Tests**
  - Added coverage for default, custom, open-vocabulary, and ontology-controlled relationship extraction.


graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -43,6 +43,76 @@
...
+# This list is deliberately domain-neutral and pairs with DEFAULT_ENTITY_TYPES.
+# It is *guidance*, not a filter: exactly like ``entity_types``, it steers the
+# prompt, and a relation the model labels outside this list is still kept. The
+# hard-filtering path is ``Ontology.relations``, which prunes non-conforming
+# edges in ``IngestionPipeline._prune``. Users who want that stricter behaviour
+# should declare an ontology; users who want none of it can pass
+# ``relation_types=[]``.
+DEFAULT_RELATION_TYPES: list[str] = [
+    # structure / place
+    "located_in",
```
galshubeli: **The list is `lower_snake_case`, but `rel_type` is UPPER_SNAKE everywhere else in the repo.**

Every other statement of the convention disagrees with this list:

- `docs/graph-schema.md:123` — ``rel_type | Original relationship type (e.g., "WORKS_AT", "LOCATED_IN")``
- `docs/graph-schema.md:146` — a copy-paste query: `MATCH (a:Person)-[r:RELATES {rel_type: "WORKS_AT"}]->(b:Organization)`
- `docs/architecture.md:71`, `docs/extraction.md:285`, `docs/strategies.md:324` — same
- `Relation` docstring, `core/models.py:263` — `label="WORKS_AT"`
- the discovery system prompt, `discovery/prompts.py:32` — `'Steve Jobs founded Apple' is FOUNDED`
- and the other half of this same function: `_relationship_type_instruction([])` still returns *"a descriptive relationship label in UPPER_SNAKE_CASE (e.g. WORKS_AT, MARRIED_TO, LOCATED_IN, CREATED_BY, PART_OF)"*

`rel_type` is a string property, so comparison is case-sensitive both in Cypher and in `_prune` (`ingestion/pipeline.py:428` compares `r.properties["rel_type"]` against `rt.label` verbatim). After this merges, on the default path: the documented example query returns zero rows; `relation_types=[]` and the default produce different casing for the same predicate; and a graph that later gains a discovered ontology ends up carrying both `located_in` (old edges) and `LOCATED_IN` (new ones) as distinct predicate strings. `_aggregate_relations` casefolds its dedup key, which hides this within a single run but not across runs.

Uppercasing the 31 labels costs nothing and keeps the finding intact — your own data says the gain is consistency, not the specific words (0.1490 vs 0.1456 against the gold vocabulary itself).

One caveat worth stating plainly: gold's predicates are lowercase (`contains`, `authored`), so if the scorer compares predicates case-sensitively, uppercasing will *look* like a regression on that harness. That's a harness artifact — casefold predicates in the comparison, then uppercase the shipped list.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -43,6 +43,76 @@
...
+# triples; ``authored``, 54) were emitted zero times.
+#
+# Supplying any fixed list doubles exact triple F1 (0.0725 -> 0.1490, +2.06x),
+# and a list written *without* reference to the gold vocabulary scored as well
+# as the gold vocabulary itself (0.1490 vs 0.1456). The gain comes from being
+# consistent, not from guessing the right words — which is what makes a shipped
+# default worth having.
+#
+# This list is deliberately domain-neutral and pairs with DEFAULT_ENTITY_TYPES.
+# It is *guidance*, not a filter: exactly like ``entity_types``, it steers the
```
galshubeli: **"guidance, not a filter" is contradicted by the prompt this default actually produces.**

Mechanically you're right — I confirmed `_prune` keys off `ontology.relations` only, and nothing prunes on `self.relation_types`. But here is the rendered prompt on the default path:

```
## Allowed Relationships
- located_in
- part_of
...
- type: MUST be one of the relationship types listed in the Allowed Relationships section above.
```

The section is headed *Allowed* and the instruction says *MUST*. Nothing rejects an off-list label, but the model is being told not to emit one — so the practical effect on a domain corpus with no declared ontology is real information loss: `acquired` → `owns`, `inhibits` → `uses`, `overrules` → nothing in the list at all. The five unseen domains were Wikipedia, which is the favorable case for a generic 31-label vocabulary.

Two ways out, either is fine:

1. Give the default path its own instruction that matches the documented semantics — *"prefer one of the labels above; if none fits, use a descriptive UPPER_SNAKE_CASE label"*. Keeps the consistency gain, stops suppressing domain predicates, and incidentally fixes the casing gap for off-list labels.
2. Drop the "guidance, not a filter" framing from this comment and the `relation_types` docstring, and say instead that the model is instructed to restrict itself but nothing enforces it.

As written, a user reading the docstring will be surprised by what the model does.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/graph_extraction.py
```diff
@@ -449,6 +529,13 @@ class GraphExtraction(ExtractionStrategy):
         coref_resolver: Optional coreference resolver applied per-chunk.
         entity_types: Entity type labels. Default: DEFAULT_ENTITY_TYPES.
             Overridden by ontology.entities if present.
+        relation_types: Relation labels offered to the LLM. Default:
```
galshubeli, line 558: **Docs and CHANGELOG not updated.**

`relation_types` is a new public kwarg, and `docs/api-reference.md:597` still reads:

```
GraphExtraction(llm, *, entity_extractor=None, coref_resolver=None, entity_types=None, max_concurrency=None)
```

This also changes the default extraction output for every user without a declared ontology, which belongs in `CHANGELOG.md`. And `docs/ontology-discovery.md:13` documents the `DEFAULT_ENTITY_TYPES` seeding by name and lists its members — the relation counterpart deserves the same line, especially since the two now behave differently (entity labels get seeded into the ontology graph, relation labels don't).

graphrag_sdk/tests/test_graph_extraction.py
```diff
@@ -1038,3 +1040,60 @@ def test_ontology_labels_are_checked_too(self):
     def test_store_and_extractor_share_one_definition(self):
         """Two hardcoded copies would drift; the bug returns when they do."""
         assert GraphStore._STRUCTURAL_LABELS is RESERVED_NODE_LABELS
+
+
+class TestDefaultRelationTypes:
```
galshubeli, line 1355: **The new branch in `extract()` has no test.**

All seven tests here are constructor- or helper-level. Neither of the two things the PR actually claims is covered: that the default vocabulary reaches a real step-2 prompt when no ontology is declared, and that `ontology.relations` overrides `self.relation_types`.

`test_relation_patterns_in_prompt` (line 742) already has the `CaptureLLM` template — two more tests off it would cover the `if ontology.relations:` fork end to end, which is where the behaviour actually lives.

Separately: `test_default_list_reaches_the_prompt` asserts `"MUST be one of"`, which locks in the prompt wording that the module comment contradicts (see my other comment). Whichever way that resolves, this assertion should move with it.

galshubeli (commented):
Verified locally on `9edad5d`, diffed against #320's head so this PR's changes are isolated: **1180 passed, 40 skipped**, and `ruff check src/` + `ruff format --check src/` clean (CI's scope). The claims in the description hold up.

The core idea is right and the measurement discipline in the comments is the best thing here. Four findings inline, plus two minor follow-ups:

- `relation_types` isn't validated, while `entity_types` goes through `_reject_reserved_labels` — a blank or whitespace label renders as a bare `- ` bullet in the Allowed Relationships list. Odd asymmetry given the "exact counterpart of `entity_types`" framing.
- Follow-up issue, not this PR: `_ensure_ontology_initialized` (`api/main.py:397`) seeds `DEFAULT_ENTITY_TYPES` specifically "so the ontology graph accurately reflects the labels the extractor will produce". That's now half-true — relations have a known default vocabulary that never reaches the ontology, so `cypher_generation.py:126`'s ``Allowed `rel_type` values:`` line stays empty and text-to-Cypher keeps guessing predicate strings the extractor now emits predictably. Seeding them isn't the fix (it would activate `_prune` and turn guidance into a filter), but surfacing the vocabulary to the Cypher schema block is worth an issue.

galshubeli (approved):

PR #322, opened 2026-09-07
Extraction: SpacyExtractor and CompositeExtractor as opt-in NER backends

Naseem77:
## Extraction — alternative NER backends (opt-in)

Part of FalkorDB/research#88. Stacked on #321. New API surface only; nothing changes unless you construct these.

### 1. `SpacyExtractor`
A second `EntityExtractor` backed by spaCy's NER (PERSON / ORG / GPE / LOC / FAC → Person / Organization / Location). Fast (~4 s for the corpus), offline. Requires `graphrag-sdk[spacy]` and a spaCy model.

### 2. `CompositeExtractor`
Chains extractors: the first one's entities win where spans overlap, later ones fill gaps. Built for GLiNER-first, spaCy-fills-in: the GLiNER default misses names like *Paris Observatory* and *Constantinople*; spaCy recovers 18 of those 24. Measured entity F1 0.568 → 0.609 at +0 s, answer accuracy unchanged — hence opt-in.

### 3. Exports
`SpacyExtractor`, `CompositeExtractor` and `DEFAULT_RELATION_TYPES` are importable from `graphrag_sdk` directly.

### Verification
Full suite: **1179 passed, 41 skipped**; ruff clean.


graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -824,3 +825,299 @@ def _parse_response(
...
+                key = ent.name.strip().casefold()
+                if not key:
+                    continue
+                existing = merged.get(key)
+                if existing is None:
+                    spans = self._spans_of(ent)
+                    if (
+                        self._suppress_overlaps
+                        and spans
+                        and any(s < ce and cs < e for (s, e) in spans for (cs, ce) in claimed)
```
galshubeli: **Overlap suppression can drop the longer, better span.**

The predicate drops a later extractor's entity on *any* overlap, with no length or containment comparison — so whichever extractor came first keeps its span even when it's the fragment. The docstring justifies this with "spaCy contributed `Fresnel` next to GLiNER's `Fresnel lens`", which assumes the earlier extractor always holds the longer span; nothing enforces that.

Concrete case, from this PR's own motivating example: GLiNER emits `Paris` (Location, span 0–5) rather than missing the phrase entirely, spaCy emits `Paris Observatory` (FAC→Location, span 0–17). `0 < 5 and 0 < 17` → the correct longer entity is discarded and the fragment kept. That inverts the stated purpose (recovering `Paris Observatory`, `Samarkand Expedition of 892`) in exactly the mixed case, and it's invisible — no log line, no counter.

Suggest preferring the longer (or containing) span over the earlier one when spans nest, and keeping the first-wins rule only for genuine partial overlaps.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -824,3 +825,299 @@ def _parse_response(
...
+        self._confidence = confidence
+        self._nlp: Any = None
+        self._lock = asyncio.Lock()
+
+    async def _get_nlp(self) -> Any:
+        if self._nlp is not None:
+            return self._nlp
+        async with self._lock:
+            if self._nlp is None:
+                self._nlp = await asyncio.to_thread(self._load)
```
galshubeli: **Per-instance model load re-introduces the memory bug `GLiNERExtractor` was fixed for.**

`_nlp` is a per-instance attribute and `_load` calls `spacy.load` unconditionally. Compare `GLiNERExtractor._MODEL_CACHE` / `_get_shared_model` (lines 566–606), whose comment records: *"Bug #11: nothing cached the model, so every `GLiNERExtractor` instance loaded its own copy … ~395 MB per copy … projects to ~11.6 GB for 30 concurrent documents. That matches the customer report of large RSS under concurrent ingest."*

`en_core_web_lg` is several hundred MB resident (dominated by 685k × 300-dim vectors). Extractors are already constructed ad hoc — `discovery/pipeline.py:773` does `GLiNERExtractor()`, `graph_extraction.py` defaults one per `GraphExtraction` — so a user following the documented `CompositeExtractor([GLiNERExtractor(), SpacyExtractor()])` pattern with one pipeline per document pays the full spaCy model cost per document while the GLiNER half stays shared.

A class-level cache keyed on `model_name`, mirroring `_MODEL_CACHE`, costs almost nothing here.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -824,3 +825,299 @@ def _parse_response(
...
+            *(e.extract_entities(text, entity_types, source_chunk_id) for e in self._extractors),
+            return_exceptions=True,
+        )
+
+        merged: dict[str, ExtractedEntity] = {}
+        claimed: list[tuple[int, int]] = []
+        failures: list[BaseException] = []
+        for extractor, result in zip(self._extractors, results, strict=True):
+            if isinstance(result, BaseException):
+                failures.append(result)
```
galshubeli, line 1339: **Configuration errors are demoted to per-chunk warnings.**

Any sub-extractor exception is logged and skipped; the composite only raises when *every* extractor fails. But `SpacyExtractor._load` raises `ImportError` ("requires the 'spacy' extra") and `OSError` ("model … is not installed. Run: python -m spacy download …") — both permanent, deterministic configuration faults.

Scenario: a user copies the docstring example without running `python -m spacy download en_core_web_lg`. The ingest completes successfully, output byte-identical to GLiNER alone, plus 157 identical WARNING lines buried in the log — and they conclude the composite bought them no recall. `_get_nlp` also never memoizes the failure, so every one of the 157 chunks re-attempts the model load in a worker thread.

Suggest failing fast on `ImportError`/`OSError` from the first chunk, or at minimum logging once per extractor rather than once per chunk.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -824,3 +825,299 @@ def _parse_response(
...
+                logger.warning(
+                    "%s failed during entity extraction, skipping its results: %s",
+                    type(extractor).__name__,
+                    result,
+                )
+                continue
+            fresh: list[tuple[int, int]] = []
+            for ent in result:
+                key = ent.name.strip().casefold()
```
galshubeli, line 1348: **Name-keyed merge discards the spans of every repeat occurrence.**

When a second entity with the same normalised name arrives, the merge branch copies `type`, `description` and `source_chunk_ids` — but never `spans`, and its spans never reach `fresh`/`claimed`. Two consequences:

1. **Provenance loss.** `_aggregate_entities` (`graph_extraction.py:817-820`) is built to accumulate every occurrence's offsets per chunk (`existing_spans.setdefault(chunk_id, []).extend(offsets)`). Feed it a `CompositeExtractor` and a name occurring three times in a chunk yields offsets for only the first — a silent span-coverage regression relative to using the same extractor directly.
2. **Suppression gap.** Because occurrences 2+ never reach `claimed`, a later extractor's fragment overlapping the *second* mention of "Fresnel lens" isn't suppressed — precisely the false positive `suppress_overlaps` exists to prevent.

graphrag_sdk/src/graphrag_sdk/ingestion/extraction_strategies/entity_extractors.py
```diff
@@ -824,3 +825,299 @@ def _parse_response(
...
+        "LANGUAGE": ("Language",),
+    }
+
+    DEFAULT_MODEL = "en_core_web_lg"
+
+    def __init__(
+        self,
+        model_name: str | None = None,
+        labels: Iterable[str] | None = None,
+        confidence: float = 0.5,
```
galshubeli: **`0.5` is not "the default GLiNER threshold", and the `confidence` arg has no behavioural effect.**

The docstring below says this "is set at the default GLiNER threshold so these entities are neither favoured nor penalised relative to GLiNER's". `GLiNERExtractor.DEFAULT_MODEL` is `urchade/gliner_medium-v2.1`, whose `DEFAULT_THRESHOLDS` entry is **0.75**; `0.5` is `_FALLBACK_THRESHOLD` / the bi-encoder value. So spaCy entities land systematically *below* GLiNER's on the default model.

That isn't cosmetic — the value is written through `_merge_step1_metadata` (`graph_extraction.py:667-669`) and `_optional_extras` (`:333-335`) onto the graph node's `confidence` property. A user filtering `confidence >= 0.75`, the natural cut given the default extractor's threshold, loses every spaCy-sourced entity.

Separately: `extract_entities` passes `self._confidence` as both the per-prediction `score` and the `threshold` into `_parse_predictions`, so `confidence < threshold` is never true and the demotion branch is structurally unreachable. The parameter has no effect beyond the number recorded on the node.

galshubeli (commented):
Reviewed the two new extractors against the PR head. The measurement work behind these is genuinely useful — the label-set ablation table in particular is the kind of thing that should be in a docstring. Six things below; #1 and the missing tests are the ones I'd want resolved before merge.

One finding has no anchor in this diff, so it goes here:

**`graph_extraction.py:628` — wrapping an `LLMExtractor` in a `CompositeExtractor` silently drops the batched LLM path.** The dispatch is `isinstance(self.entity_extractor, LLMExtractor)`, which is False for a composite, so `CompositeExtractor([LLMExtractor(llm), SpacyExtractor()])` takes the local branch. Three effects: each chunk issues its own `ainvoke` instead of a single `abatch_invoke`; the caller's `max_concurrency` no longer reaches the LLM client; and `LLMExtractor.extract_entities` formats the prompt with `", ".join(entity_types)` rather than `_format_entity_types(entity_types, entity_type_descs)`, so ontology entity-type descriptions vanish from the NER prompt and extraction quality changes. The composite's sub-extractors also run concurrently *inside* the per-chunk `Semaphore(max_concurrency or 12)`, so effective outbound concurrency is N× the configured limit. Worth either documenting as unsupported or rejecting `LLMExtractor` members in `CompositeExtractor.__init__`.

**No tests.** The diff adds two public classes and touches only `entity_extractors.py` plus two `__init__.py` files; `tests/test_entity_extractors.py` is untouched. Nothing exercises `_type_for`, the label mapping, overlap suppression, the merge ordering rule, or the all-fail re-raise — which is why the inline findings below pass CI cleanly.

Two things I checked and am **not** flagging: spaCy thread safety is fine (12 threads sharing one `nlp` over 400 docs, matching the pipeline's `Semaphore(12)` + `asyncio.to_thread` shape — 0 mismatches), and the `spacy = ["spacy>=3.0"]` extra already exists in `pyproject.toml:60`, so there's no packaging gap.

Nit: `disable=["lemmatizer", "textcat"]` — `textcat` isn't in any `en_core_web_*` pipeline, so it's silently ignored (dead but harmless), and the adjacent comment says "the parser and lemmatizer cost time" while the parser isn't actually disabled.

galshubeli (approved):

PR #317, opened 2026-09-06
Resolution and dedup: LLM-verified default, unified candidate stage, finalize-time fixes

Naseem77:
## Summary

Reworks entity resolution and finalize-time deduplication. Every default here was measured against a real FalkorDB and a real LLM on the benchmark gold set (216 duplicate pairs, 39 must-not-merge pairs); numbers are in the commit messages and CHANGELOG.

## Commits

1. **refactor(resolution)** — remove `SemanticResolution` and `DescriptionMergeResolution`. Neither was reachable from production code; both were strictly weaker than `LLMVerifiedResolution`.
2. **feat(resolution)** — `LLMVerifiedResolution`: unified same/cross-label candidate stage (`unified_threshold=0.65`, `soft_threshold` 0.80→0.65: +26 % recall, 0 extra wrong merges), batched verification on by default (5 pairs/call, 4000-token budget — larger batches were bimodal on identical input), `merged_labels` recorded on cross-label merges, `flatten_remap` for merge chains. `ExactMatchResolution` now matches on `name` (was `id`, which never merged anything).
3. **feat(storage)** — `merged_labels` promoted to real Cypher labels on write (a property is not a type: `MATCH (n:Engineer)` missed nodes that had absorbed `Engineer`). Structural labels are never promoted.
4. **fix(dedup)** — `EntityDeduplicator`: survivor is `MATCH`ed before every `MERGE` (path-MERGE created 136 NULL-name stubs from 24 merges), `source_chunk_ids` unioned, absorbed descriptions kept, safer name normalisation (`A.I.`→`AI`, foreign articles kept), acronym fold with same-label and ambiguity guards (`US` matching two long forms is left alone). Grouping stays `(name, label)`.
5. **feat(api)** — `ingest()`/`update()` default resolver is `LLMVerifiedResolution`; pass `resolver=ExactMatchResolution()` for a zero-cost ingest. Fuzzy dedup threshold 0.9→0.95 (0.90 merged 2 of 33 hard negatives for no recall).
6. **feat(api)** — `retrieve()` warns once when documents were ingested without a `finalize()`/`deduplicate_entities()` pass.
7. **docs** — ingestion, strategies, api-reference, storage pages and CHANGELOG updated to match.

## Breaking

- `SemanticResolution`, `DescriptionMergeResolution` removed.
- `ingest()` default resolver now makes embedding and LLM calls. Opt out with `resolver=ExactMatchResolution()`.
[...]

PR #323, opened 2026-09-07
Deduplication: finalize-time deduplicator fixes — ghost nodes, kept descriptions, name normalisation, acronyms

Naseem77:
## Deduplication — finalize-time deduplicator fixes

Measured on an 11-document benchmark against a real FalkorDB. Part of FalkorDB/research#88. First of five stacked PRs replacing #317.

### 1. Merging created empty ghost nodes
**Problem:** the edge-remap Cypher used a path-`MERGE` with an unbound survivor `(s {id})`, which creates a nameless `__Entity__` stub for every edge the survivor did not already have — 136 stubs from 24 merges on a 445-entity graph. The remapped edges hung off the stubs until `finalize()`'s NULL cleanup deleted them, losing the facts.
**Fix:** the survivor is `MATCH`ed before every `MERGE`; RELATES remaps union `source_chunk_ids` instead of overwriting. 0 stubs.

### 2. The absorbed entity's description died with it
**Problem:** only the survivor's description was kept.
**Fix:** descriptions are joined with `" | "` and the absorbed names recorded in `aliases`.

### 3. Name grouping was too loose in some places and too strict in others
**Fix:** normalisation folds accents, punctuation and inner dots (`A.I.` → `AI`) and strips only a leading **English** article (`Los Angeles` and `Le Mans` stay intact); generational suffixes are kept (`Elias Whitford, Jr.` ≠ `Elias Whitford`). An acronym joins its long form when exactly one same-label candidate spells it (`AIHS` / `Ashford Island Historical Society`); an ambiguous one (`US`) is left alone. Names in a non-Latin script no longer fold to `""` and merge with each other.

### 4. Fuzzy threshold 0.90 → 0.95
**Problem:** at 0.90 the name-embedding tier merged 2 of 33 deliberate hard-negative pairs for no added recall.
**Fix:** 0.95 in `EntityDeduplicator` and `deduplicate_entities()`.

### 5. "Index already exists" logged as ERROR
**Problem:** `ensure_indices()` re-creates indexes idempotently; FalkorDB answers with a non-transient error, surfaced as 5 ERROR lines per `finalize()`.
**Fix:** recognised and logged at DEBUG; every other non-transient failure keeps ERROR.

### Verification
Full suite: **1148 passed, 41 skipped**; ruff clean.

## Summary by CodeRabbit

- **New Features**
  - Added configurable relation types and open-vocabulary extraction options.
  - Improved entity deduplication with multilingual normalization, alias and provenance merging, safer edge remapping, and a higher fuzzy-match threshold.
  - Added adaptive GLiNER extraction settings and overlapping-window processing.

- **Bug Fixes**
  - Re-ingesting unchanged documents is now a true no-op; forced updates reprocess them.
  - Prevented duplicate chunks during repeated ingestion.
  - Expected database index-exists errors are handled without unnecessary error logging.

- **Changed**
  - Reduced the default chunk size to 384 tokens.
  - Improved relation extraction to avoid stopping prematurely.

- **Documentation**
  - Updated API, configuration, and storage documentation for these changes.


graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -16,6 +18,65 @@
...
+    correctness bug rather than a cleanup.
+    """
+    original = re.sub(r"\s+", " ", str(name or "")).strip()
+    s = unicodedata.normalize("NFKD", original).encode("ascii", "ignore").decode()
+    s = s.lower().strip()
+    s = re.sub(r"(?<=\w)\.(?=\w|$)", "", s)
+    s = re.sub(r"[^a-z0-9 ]", " ", s)
+    s = re.sub(r"\s+", " ", s).strip()
+    cleaned = _LEADING_ARTICLE.sub("", s).strip() or s
+    if cleaned:
```
galshubeli: **Mixed-script names collapse to their ASCII remnant and get merged.**

The non-Latin guard only fires when the ASCII fold is *entirely* empty. One surviving ASCII token is enough to defeat it, and then the token alone becomes the group key:

| names (same label) | normalises to |
|---|---|
| `Отдел 5`, `Кабинет 5` | `5` |
| `東京, Japan`, `大阪, Japan` | `japan` |
| `Школа №1`, `Гимназия №1` | `no1` (NFKD maps `№`→`No`) |

Reproduced end-to-end: feeding `["Отдел 5", "Кабинет 5"]` (both `Location`) to `deduplicate()` returns `merged=1` and issues `DETACH DELETE {dup_id: e1}`. The old key (`name.strip().lower()`) never did this, so it's a new data-loss path on any non-English corpus.

Suggest comparing the fold against the original rather than only testing for empty — fall back to `original.casefold()` whenever non-ASCII characters were stripped:

```python
if cleaned and not any(ord(ch) > 127 for ch in original):
    return cleaned
return original.casefold()
```

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -16,6 +18,65 @@
...
+    Society``: their character similarity is 0.11 and their embedding similarity
+    sits below every threshold worth using, so neither fuzzy matching nor
+    vectors recover that pair.
+
+    Kept strict on purpose (2-6 alphabetic characters; the long form must have
+    at least two significant words). A looser version merges arbitrary
+    unrelated short strings.
+    """
+    s = normalize_entity_name(short).replace(" ", "")
+    if not (2 <= len(s) <= 6) or not s.isalpha():
```
galshubeli: **The 2-character lower bound merges unrelated people.**

With `2 <= len(s)`, any two-letter name spells the initials of any same-label two-word name. Confirmed: `Bo`/`Person` and `Ben Ottoson`/`Person` produce `merged=1` — one node is `DETACH DELETE`d with no LLM in the loop. Likewise `Al`/`Anna Lindqvist`, `Jo`/`Jane Oliver`.

The ambiguity guard in `_merge_acronym_groups` doesn't help here: in a graph with a single two-word `Person` whose initials match, `len(matches) == 1` and the union proceeds.

Two-letter strings carry too little signal to be acronym evidence. Either raise the floor to `3 <= len(s)`, or require ≥3 significant words in the long form when the short form is 2 characters.

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -126,12 +221,95 @@ async def _deduplicate_exact(self, batch_size: int) -> int:
...
+                parent[rb] = ra
+
+        names = [g[0]["name"] for g in groups]
+        labels = [{(e.get("label") or "").strip().lower() for e in g} - {""} for g in groups]
+        shorts = [
+            i for i, n in enumerate(names) if len(normalize_entity_name(n).replace(" ", "")) <= 6
+        ]
+        longs = [i for i, n in enumerate(names) if len(normalize_entity_name(n).split()) >= 2]
+        for i in shorts:
+            matches = [
```
galshubeli: **The acronym fold is O(shorts × longs) with the normalisation re-run per pair.**

For every candidate pair, `is_acronym_of` re-runs `normalize_entity_name` on both names and `_initials(long)` normalises the long form a third time — five regex passes plus an NFKD encode per call. Measured ~51 µs/pair; 400k pairs took 20.4 s.

`labels[i] & labels[j]` prunes very little when the ontology has only a handful of labels, so on a large graph this is effectively `len(shorts) × len(longs)`. That's a quadratic cost inside `finalize()`, in the same code path this PR is otherwise speeding up.

Since initials are a pure function of the name, this can be O(n): normalise each group's name once, bucket `longs` by initials, then do one dict lookup per short.

```python
norm = [normalize_entity_name(n) for n in names]
by_initials = defaultdict(list)
for j in longs:
    by_initials[_initials_from_norm(norm[j])].append(j)
for i in shorts:
    matches = [j for j in by_initials.get(norm[i].replace(" ", ""), [])
               if i != j and labels[i] & labels[j]]
```

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -102,12 +193,14 @@ async def _deduplicate_exact(self, batch_size: int) -> int:
         # Group by (normalized name, label) to prevent cross-type merging.
         groups: dict[tuple[str, str], list[dict]] = {}
         for ent in entities:
-            norm = ent["name"].strip().lower()
+            norm = normalize_entity_name(ent["name"])
             label = ent.get("label", "").strip().lower()
             groups.setdefault((norm, label), []).append(ent)
 
+        merged_groups = self._merge_acronym_groups(list(groups.values()))
```
galshubeli, line 278: **After an acronym union, "survivor = longest description" can keep the acronym and delete the long form.**

The sort just below (`group.sort(key=lambda e: len(e["description"]))`) is description-length only, so it has no idea one of these names is an acronym of another. Confirmed: with `AIHS` carrying the longer description, `deduplicate()` issues `DETACH DELETE` on `Ashford Island Historical Society` and the graph keeps `AIHS`.

That's worse than a cosmetic loss, because `backfill_entity_embeddings()` embeds the **name only** — `vector_store.py:678` collects `name` and discards `desc` — so the surviving entity is vector-indexed as the opaque string `AIHS`, which degrades entity retrieval for exactly the queries that used to hit the long form.

When a group was unioned via `is_acronym_of`, the long form should win regardless of description length.

Separately: the PR description says "the absorbed names recorded in `aliases`", but there's no `aliases` code in the diff — so the acronym↔long-form association is lost entirely. Worth adding it, or dropping that line from the description.

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -126,12 +221,95 @@ async def _deduplicate_exact(self, batch_size: int) -> int:
...
+
         logger.info(f"EntityDeduplicator phase 1 (exact): merged {merged} duplicates")
         return merged
 
+    async def _merge_descriptions(self, survivor: dict, absorbed: list[dict]) -> None:
+        """Fold the absorbed duplicates' descriptions into the survivor."""
+        descriptions: list[str] = []
+        for ent in [survivor, *absorbed]:
+            desc = str(ent.get("description") or "").strip()
+            if desc and desc not in descriptions:
```
galshubeli: **Description dedup compares whole strings, so segments repeat across successive merges.**

`desc not in descriptions` is exact element containment against the list, not a check on the existing `" | "` segments. Once a survivor holds `"a lighthouse | first lit in 1871"`, absorbing a node described as `"first lit in 1871"` yields:

```
a lighthouse | first lit in 1871 | first lit in 1871
```

Confirmed by running `deduplicate()` on that pair. Across incremental ingest/finalize cycles it compounds on hub entities, and these strings feed the `__Entity__(name, description)` fulltext index.

Splitting the accumulated description before the check fixes it:

```python
for ent in [survivor, *absorbed]:
    desc = str(ent.get("description") or "").strip()
    for seg in (s.strip() for s in desc.split(" | ")):
        if seg and seg not in descriptions:
            descriptions.append(seg)
```

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -26,21 +87,44 @@
...
+# ``__Entity__`` node carrying only ``id``. The remapped edge then attaches to
+# that empty duplicate instead of the real survivor, so a "deduplication" run
+# both invents entities and silently loses relationships. Reproduced minimally:
+# 3 entities in, 4 out, "Nodes created: 1", two nodes sharing one id.
 _REMAP_QUERIES = [
     # Outgoing RELATES from duplicate.
     "MATCH (dup:__Entity__ {id: $dup_id})-[r:RELATES]->(b:__Entity__) "
     "WHERE b.id <> $survivor_id "
-    "MERGE (s:__Entity__ {id: $survivor_id})-[nr:RELATES]->(b) "
+    "MATCH (s:__Entity__ {id: $survivor_id}) "
```
galshubeli, line 150: The `MERGE`→`MATCH` change is the right fix, and the comment above explaining the path-`MERGE` ghost-node behaviour is genuinely useful — but it turns a missing survivor into a silent no-op that still reports success.

If the survivor node is gone, all three remap queries match zero rows and raise nothing, so `_remap_entity_edges` returns `True` and the caller proceeds to `MATCH (e {id: $dup_id}) DETACH DELETE e` — every one of the duplicate's `RELATES` and `MENTIONED_IN` edges is destroyed rather than remapped.

Reachable because `_fetch_all_entities` snapshots the entity list up front, so a `delete_document()` or a second `finalize()` running concurrently with the dedup pass can remove a survivor mid-loop.

Cheap to close: have the remap confirm the survivor exists before returning `True` (e.g. `RETURN count(s) AS n` on the first query, or one pre-flight lookup per group).

graphrag_sdk/src/graphrag_sdk/core/connection.py
```diff
@@ -200,7 +200,13 @@ async def query(
                 last_exc = exc
                 # Don't retry non-transient errors (e.g. schema/index conflicts)
                 if self._is_non_transient(exc):
-                    logger.error(
+                    # "Already indexed / already exists" is the expected reply
+                    # to an idempotent CREATE INDEX on a graph that has one;
+                    # VectorStore treats it as success. Logging it as an error
+                    # made every finalize() print three spurious failures.
+                    level = logging.DEBUG if self._is_already_exists(exc) else logging.ERROR
```
galshubeli: **This suppression is a layer lower than the knowledge that justifies it.**

`vector_store.py:27` already defines an identical `_INDEX_EXISTS_MARKERS = ("already indexed", "already exists")`, and `_try_create_index` already matches it and returns `True`. Every index-creation call site goes through that path, so the check here adds nothing for the intended case — but `query()` can't tell an idempotent `CREATE INDEX` from any other query, so it also silences unrelated ones.

Concrete case: a graph was indexed at `embedding_dimension=256` and the user reconfigures to 1536. `create_entity_vector_index()` gets an "already indexed" reply, `_try_create_index` reports success, `ensure_indices()` reports all-green, and after this change nothing is logged above DEBUG. Vector search then queries a 256-dim index with 1536-dim vectors and there is no trace anywhere in the logs — before this PR there was at least one ERROR line.

The spurious-log problem is real; suggest fixing it at the call site that knows the query was idempotent (have `_try_create_index` pass a flag, or log there and keep `query()` honest).

galshubeli (commented):
Reviewed the five fixes. The core diagnoses hold up well — the path-`MERGE` ghost-node analysis is correct and clearly explained, and the `source_chunk_ids` union catches a real provenance bug that would have bitten `delete_stale_relationships` later. The `MATCH`-before-`MERGE` fix and the generational-suffix decision are both right.

My concern is concentrated in the new name normalisation and acronym fold: three of the findings below are data-loss paths that the old `name.strip().lower()` key did not have, and they delete nodes without an LLM in the loop.

**Needs a change before merge**
- `normalize_entity_name` collapses mixed-script names to their ASCII remnant (`Отдел 5` and `Кабинет 5` both → `5`) and merges them.
- `is_acronym_of` accepts 2-character short forms, so `Bo` merges with `Ben Ottoson`.
- The acronym fold is O(shorts × longs) with the normalisation re-run per pair (~51 µs each) — a quadratic regression inside `finalize()`, fixable to O(n) by bucketing on initials.

**Worth fixing**
- Acronym unions can keep the acronym and delete the long form, which then gets vector-indexed as an opaque string.
- Description merging repeats segments across successive merges.
- A missing survivor makes the remap a silent no-op that still reports success.
- The "already exists" downgrade sits in `query()`, where it can mask a real index/dimension conflict.

**Docs (not in this diff)**
The threshold default moved to `0.95` in both `EntityDeduplicator.deduplicate` and `GraphRAG.deduplicate_entities`, but the published docs still say `0.9`:
- `docs/configuration.md:576` — code sample
- `docs/configuration.md:586` — reference table row
- `docs/api-reference.md:177` — signature block

Findings 1, 2, 4 and 5 I reproduced end-to-end against the mock harness in this PR's own test file; 3 by microbenchmark. The remap and logging findings are reasoning from the code, not executed.

galshubeli (approved):

PR #329, opened 2026-09-09
docs: clarify multi-document ingestion concurrency

### LINKED ISSUE #286: docs(ingestion): concurrency tuning guide + warning for the sequential per-file ingest() pattern

sen-to:
## Summary

Document the document-level concurrency enabled by `ingest(list_of_sources)`.

Clarify how `GraphRAG.ingest(max_concurrency=...)`, `GraphExtraction.max_concurrency`, and provider concurrency differ.

Explain how `apply_changes()` uses `max_concurrency` for added files and why `update_concurrency` remains `1` by default.

Closes #286

## Validation

`npm install --no-save @mdx-js/mdx@3.1.1 remark-frontmatter@5.0.0 remark-gfm@4.0.1 && node scripts/check_mdx.mjs`

Result: 19 files, 0 with syntax errors.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for efficiently ingesting multiple documents concurrently.
  - Documented configurable concurrency limits and input-order result handling.
  - Added examples for tuning document-level and language-model processing concurrency.
  - Documented equivalent concurrency controls when applying changes.


galshubeli (commented):
Three notes on the concurrency semantics in the new section — the rest of the page checks out against the source (batch result ordering, the `IngestionResult | Exception` contract, the `3` / `1` defaults, and the "does not parallelize within one document" point are all accurate).

docs/ingestion.mdx
```diff
@@ -247,6 +247,48 @@ stats = await rag.finalize()
...
+Increase `max_concurrency` gradually when the document loader, LLM provider,
+and FalkorDB deployment have capacity. A higher value increases simultaneous
+document pipelines and their memory use; it does not make one document's
+pipeline run in parallel.
+
+This is separate from extraction concurrency. `GraphExtraction` limits the
+parallel LLM calls made for chunks *within* each document; the configured LLM
+provider also has its own `max_concurrency` limit. Tune document-level and
+LLM-level concurrency together, and lower either value if the provider starts
+returning rate-limit errors.
```
galshubeli, line 276: **"lower either value" doesn't hold once the extractor sets a value.**

`GraphExtraction._max_concurrency` is forwarded as `abatch_invoke(max_concurrency=...)`, and the provider resolves it with:

```python
sem = asyncio.Semaphore(max_concurrency or self.max_concurrency)  # core/providers/base.py:285
```

That's an **override**, not a `min()`. So in the configuration the next code block sets up (`GraphExtraction(llm=llm, max_concurrency=6)`), lowering `llm.max_concurrency` has no effect on extraction at all — only the extractor's value is live.

Suggest saying explicitly that the extractor's `max_concurrency` *replaces* the provider's when set, and that the provider default (12) only applies when the extractor leaves it at `None`.

docs/ingestion.mdx
```diff
@@ -247,6 +247,48 @@ stats = await rag.finalize()
...
+
+```python
+from graphrag_sdk.ingestion.extraction_strategies.graph_extraction import GraphExtraction
+
+extractor = GraphExtraction(llm=llm, max_concurrency=6)
+results = await rag.ingest(sources, extractor=extractor, max_concurrency=3)
+```
```
galshubeli, line 283: **These two numbers multiply, and the text doesn't say so.**

The semaphore inside `abatch_invoke` is constructed per call, and each document pipeline makes its own call — nothing is shared across concurrent documents. So this example allows up to **18** in-flight LLM calls (3 documents x 6 chunk-level calls), not 6.

That product is the number a reader needs to stay under a provider rate limit, and "Tune document-level and LLM-level concurrency together" is too vague to convey it. Worth stating the relationship directly, e.g. "peak concurrent LLM calls is `max_concurrency` x the extractor's `max_concurrency`".

Minor, same block: every other doc and example imports this from the short public path — `from graphrag_sdk import GraphExtraction` (see `configuration.mdx:410`, `strategies.mdx:244`, `examples/03_custom_strategies.py:34`). The deep module path works but reads as private API.

docs/ingestion.mdx
```diff
@@ -247,6 +247,48 @@ stats = await rag.finalize()
...
+
+For incremental refreshes, `apply_changes()` has the same
+`max_concurrency` control for newly added documents. Keep
+`update_concurrency` at its default of `1` unless you have verified that
+concurrent updates are safe for your workload.
```
galshubeli, line 288: **This softens a correctness warning into what reads like a throughput caveat.**

"unless you have verified that concurrent updates are safe for your workload" is materially weaker than the two existing sources of truth:

- `api/main.py:3934`: "**Default is 1, and you almost certainly should not raise it.**"
- `docs/incremental-updates.mdx:197`: same wording, plus a dedicated section.

The real condition is narrow and specific: raising it is safe **only** if no two updates running in parallel can ever share an entity in their candidate snapshots. Violating it makes orphan cleanup wrongly delete entities — silent data loss, not a slowdown. "Safe for your workload" doesn't signal that.

Suggest matching the existing phrasing and linking to [Concurrency and the `update_concurrency` trap](https://github.com/FalkorDB/GraphRAG-SDK/blob/main/docs/incremental-updates.mdx#concurrency-and-the-update_concurrency-trap) (`incremental-updates.mdx:239`) instead of paraphrasing it — the paraphrase is how the drift happened.

PR #331, opened 2026-09-10
fix: serialize concurrent GraphRAG lazy initialization

### LINKED ISSUE #330: Unguarded lazy init: concurrent requests redundantly load and re-register the ontology
## Summary

`GraphRAG._ensure_ontology_initialized()` and `_validate_graph_config()` are unguarded lazy-init blocks. Both are awaited on every read path, and neither is protected against concurrent entry, so a burst of concurrent requests against a freshly constructed `GraphRAG` instance all run the full init — including ontology **writes** — instead of one request initializing and the rest reusing the result.

This surfaced while investigating a user report of retrieval throughput on a single graph looking serialized under load.

## Details

`_ensure_ontology_initialized()` (`graphrag_sdk/src/graphrag_sdk/api/main.py:387`) is a check-then-act on a plain bool:

```python
if self._ontology_initialized:
    return
loaded = await self._ontology_store.load()
if self.ontology.entities or self.ontology.relations:
    self._global_ontology = await self._ontology_store.register(self.ontology)
...
self._ontology_initialized = True
```

The flag is only set after several awaits, so every task that enters before the first one finishes passes the guard. Per redundant entry that is:

- 5 sequential read queries in `OntologyStore.load()` (`storage/ontology_store.py:136`), plus
- a full `OntologyStore.register()` pass when a user ontology was supplied — which **writes** to the `<graph>__ontology` graph.

FalkorDB takes an exclusive per-graph lock for writes, so N concurrent cold requests serialize against each other on redundant registration of an ontology that is already correct. The writes are idempotent, so this is a performance and contention bug, not a correctness one.

`_validate_graph_config()` (`api/main.py:2888`) has the same shape, guarded by `self._config_validated`. Its cost per redundant entry is a config query plus a live embedder probe (`aembed_query("dim_check")`), i.e. a billable provider round trip per in-flight request until one completes.

Both are reached from `retrieve()` (`api/main.py:2638-2639`) and `completion()` (`api/main.py:2784`), so this is on the hot path for every query, not just ingest.

## Reproduction

```python
rag = GraphRAG(connection=cfg, llm=llm, embedder=embedder, ontology=my_ontology)
await asyncio.gather(*[rag.retrieve("q") for _ in range(50)])
```

Expected: one ontology load + register, one embedder probe.
Actual: up to 50 of each. Observable as 50x `OntologyStore.load()` query volume against the `__ontology` graph, 50 embedder calls, and write-lock contention on the first burst.

## Suggested fix

An `asyncio.Lock` per instance with a double-checked flag inside:

```python
async with self._init_lock:
    if self._ontology_initialized:
        return
    ...
```

Same for `_validate_graph_config()` (separate lock, or one shared init lock). Notes:

- The lock must be created lazily or in `__init__` without binding a loop — `asyncio.Lock()` is loop-agnostic on 3.10+, so `__init__` is fine.
- `get_ontology()` deliberately re-reads after `_ensure_ontology_initialized()` to pick up other writers; that behavior should be preserved — only the one-shot init is being serialized.
- Worth a regression test asserting a single `register()` call under `asyncio.gather` of N retrievals (the existing test suite has no concurrency coverage for this path).

This does not address per-instance sharing across event loops (the `*_sync` wrappers call `asyncio.run()` per call) — that is a separate concern.

mikemikimike:
## Summary

Concurrent cold requests on one `GraphRAG` instance could duplicate lazy ontology/configuration work, causing redundant ontology writes and repeated embedder probes under load.

## Changes

- Serialize ontology initialization and graph configuration validation per `GraphRAG` instance.
- Share one in-flight configuration validation task across concurrent callers while respecting each caller's latency budget.
- Coalesce transient validation failures with a short retry backoff instead of creating a failure convoy.
- Keep validation state generation-aware so `delete_all()` cannot be overwritten by an older validation task.
- Preserve newer ontology assignments during concurrent initialization.

## Tests

- `python -m pytest tests/test_facade.py -q` — 185 passed.
- `python -m pytest tests/ -q` — 2076 passed, 227 skipped.
- `ruff check src/` — passed.
- `git diff --check` — passed.

## Issue

Fixes #330

Original issue: https://github.com/FalkorDB/GraphRAG-SDK/issues/330


graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -332,6 +332,7 @@ def __init__(
         self.ontology = ontology or Ontology()
         self._embedding_dimension = embedding_dimension
         self._config_validated = False
+        self._config_validation_lock = asyncio.Lock()
```
galshubeli: **Lock is not rebound per event loop.**

This is a single `asyncio.Lock()` built in `__init__` and never rebound to the running loop — unlike the sibling `_ontology_lock()` accessor, which exists for exactly this reason. Its docstring states the rule:

> the `*_sync` wrappers run each call under a fresh `asyncio.run()`, and a lock that once waited on one loop raises when acquired from another.

CPython binds `Lock._loop` on the first *contended* `acquire()`; a second contended acquire from a different loop raises. Reproduced on the repo's Python 3.10.12:

```
loop1 ok, bound to <_UnixSelectorEventLoop ...>
loop2 RuntimeError: <asyncio.locks.Lock object ...> is bound to a different event loop
```

The window needs `_config_validated` to stay `False` across calls, which happens on any transient `query_raw`/probe failure (both paths `return` without latching it) and after `delete_all()`, which resets the flag.

Concrete failure: the embedder is briefly unreachable; an `asyncio.run(...)` that gathers two `retrieve()` calls contends the lock and binds it to loop A; a later `retrieve_sync()` / `ingest_sync()` run that contends it again raises `RuntimeError` out of `_validate_graph_config` — turning a recoverable probe failure into a hard crash.

Fix: route this through the same loop-rebinding accessor pattern as `_ontology_lock()`.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -4407,85 +4409,88 @@ async def _validate_graph_config(self, *, ctx: Context | None = None) -> None:
         """
         if self._config_validated:
             return
+        async with self._config_validation_lock:
```
galshubeli: **Lock convoys all callers behind a failing network probe.**

The lock is held across both `query_raw` and the embedder probe, and both failure paths `return` without setting `_config_validated`. So under a sustained provider or graph outage the critical section never latches, and the single-flight becomes a permanent serialization point rather than a one-time de-duplication.

With the embedder timing out, 20 concurrent `retrieve()` calls previously each paid one probe timeout in parallel; now caller N waits N×timeout, because each waiter re-runs the whole probe after the previous one returns without latching.

There is also no latency-budget escape from the queue: the four internal call sites at lines 2010, 2403, 2506 and 3347 pass no `ctx`, so the `ctx.ensure_budget(...)` guards inside are skipped and requests pile up until the caller-side timeout.

A shared in-flight future — or caching the failure for a short backoff window — would give the intended de-duplication without turning an outage into serialized latency amplification.

galshubeli (commented):
Reviewed the concurrency changes. The `set_ontology` fix looks correct — I walked the interleavings of `set_ontology` against an in-flight `_ensure_ontology_initialized` in both orders, and the double-check on `_ontology_initialized` after acquiring makes the newer assignment win. Switching to `_initialize_ontology()` (the lock-free body) correctly avoids self-deadlock on the non-reentrant lock, and there is no re-entrancy into `_ontology_lock()` via `_register_structured_ontology_locked` → `_ontology_store.register`. The `_validate_graph_config` reindentation preserves the `try/except/else` structure and keeps `_config_validated = True` inside the `async with`.

Two issues on the new config-validation lock, inline below. The first is the one I would block on.

One note on the description: the double-checked `_ensure_ontology_initialized` and `_ontology_lock()` already exist on `main`, so this PR's only ontology change is bringing `set_ontology` under that lock — `test_ontology_initialization_is_single_flight` covers pre-existing behavior. Not a defect, just worth correcting in the summary.

I could not run the suite locally (no virtualenv in my clone), so both findings are static analysis plus a standalone `asyncio` repro.

PR #333, opened 2026-09-16
Record what the judge decided about each pair

Naseem77:
## The problem

The judge reports how a run went — how many candidates it saw, how often the two passes agreed, how many merged. All of those are properties of the run across the whole graph.

None of them answers the question a reader actually has when one proposed merge is put in front of them: **how sure are we about these two**. And a run figure shown next to a single pair reads as though it were about that pair. A caller building a review UI on `judge_stats` today has nothing else to reach for, so that misreading is the likely outcome rather than an unlikely one.

## What this adds

`FinalizeResult.judge_pair_decisions` — one entry per pair the judge actually considered:

| field | meaning |
|---|---|
| `a`, `b`, `a_name`, `b_name` | the two entities |
| `similarity` | what the gates measured, before the model was asked anything |
| `votes` / `passes` | how many passes called them the same, out of how many saw them |
| `verdict` | `same`, `split`, `different`, `unjudged` |
| `merged` | whether they ended up merged |

Only pairs that reached the judge appear. A pair no gate brought forward was never a question.

## What it deliberately does not add

**No probability, no score, no confidence band.** The judge does not produce one, and deriving a number here would put a confidence on the record that nothing measured. The parts are reported instead, so a caller that wants bands can form them and will know what its own bands mean. Banding is a presentation decision and belongs to whoever is presenting.

## Details worth a second look

- `votes` and `merged` are separate, because agreement is not merger — a pair can be agreed and still be kept apart by the grouping rules.
- A pair in a set whose prompt failed is `unjudged`, not a disagreement. It was judged fewer times than the run was configured for, so "1 of 2 passes" would report a split that never happened.
- The list is cleared at the **start** of a run, not the end. Every early return otherwise leaves the previous run's decisions attributed to this one.
- Capped at 200 and ordered by similarity, so a graph with thousands of near-misses cannot turn one field into the whole payload, and the cut falls on the least similar.
- `same2 = {}` in the single-pass branch is new support for the per-pair record, not a fix to existing behaviour — it was previously bound and read only under `vote=True`.

## Tests

Eight new cases in `test_judge_dedup.py`:

- `test_pair_decisions_are_recorded_for_the_pair_that_merged`
- `test_a_pair_the_passes_split_on_is_not_reported_as_agreed`
- `test_a_single_pass_run_never_claims_two_votes`
- `test_a_rejected_pair_is_recorded_as_different`
- `test_decisions_are_ordered_by_similarity`
- `test_a_run_with_no_candidates_records_no_decisions`
- `test_decisions_do_not_survive_into_the_next_run`
- `test_protected_pairs_are_never_offered_as_decisions`

Full SDK unit suite: **2076 passed, 226 skipped, 0 failed.** Each of the two commits is green on its own.

## Summary by CodeRabbit

- **New Features**
  - Finalization results now include per-pair entity matching decisions.
  - Each decision includes entity names, similarity, nomination source, voting details, verdict, and whether the pair was merged.
  - Decisions prioritize merged pairs, are capped at 200 entries, and use similarity for subsequent ordering.
  - Results distinguish confirmed matches, disagreements, rejected pairs, and pairs that could not be evaluated.
  - Decisions remain available after partial processing failures and are empty when judging does not run.

- **Documentation**
  - Added reference documentation for per-pair matching decisions.


graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -893,5 +904,72 @@ def protected_idx(i: int, j: int) -> bool:
...
+        never produced; the caller can band them, and knows what the bands mean
+        because the parts are here.
+
+        Only pairs the judge actually considered appear. A pair no gate brought
+        forward was never a question.
+        """
+        passes = 2 if self._vote else 1
+        decisions: list[dict[str, Any]] = []
+        for (a, b), similarity in sorted(edges.items(), key=lambda kv: -kv[1]):
+            set_id = agreed.get((a, b), disagreed.get((a, b), same1.get((a, b), same2.get((a, b)))))
```
galshubeli: **A pair the judge never saw is reported as `different`.**

`set_id` is looked up only in `agreed` / `disagreed` / `same1` / `same2`. But when an LLM call fails (`not item.ok`) or a set comes back unanswered, `_judge` adds the set to `failed_sets` and records **no pairs at all** in `same`. So `set_id` is `None`, `unjudged` is `False`, `votes` is `0`, and the verdict falls all the way through to `different` with `passes: 2`.

That is the opposite of what `unanswered_sets` was written for — its docstring says the caller "treats these as unjudged rather than as 'nothing is the same'".

Reproduced against the existing `ROWS` fixture with `ScriptedLLM([None, None])` (every call fails in both passes):

```
STATS {'entities': 3, 'candidates': 3, 'sets': 1, 'llm_calls': 2, 'llm_failed': 2, 'agreed_pairs': 0, ...}

{"a": "e1", "b": "e2", "similarity": 0.9536,
 "votes": 0, "passes": 2, "verdict": "different", "merged": false}
```

The judge was never asked anything, and the record says it looked twice and said no. `ScriptedLLM([[], None])` (pass 1 answers, pass 2 fails) is the same shape: reported as `passes: 2` though only one pass saw the pair.

A review UI built on this field will tell an operator the judge rejected pairs it never saw — which is worse than showing them a run-level figure, the thing this PR exists to fix.

The fix is to pass the judged `sets` / `set_ids` (or a pair→set map built from them) into `_pair_decisions`, so set membership is known independently of what the model answered.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -893,5 +904,72 @@ def protected_idx(i: int, j: int) -> bool:
...
+        single number from those would be inventing a probability the judge
+        never produced; the caller can band them, and knows what the bands mean
+        because the parts are here.
+
+        Only pairs the judge actually considered appear. A pair no gate brought
+        forward was never a question.
+        """
+        passes = 2 if self._vote else 1
+        decisions: list[dict[str, Any]] = []
+        for (a, b), similarity in sorted(edges.items(), key=lambda kv: -kv[1]):
```
galshubeli: **A pair that merged can be missing from the record entirely.**

This iterates `edges` — the gate-nominated pairs. But the judge is shown whole sets and asked about *every* pair inside one, and `make_groups_dense` admits a set at >50% density. So a set can contain a pair that was never an edge, and `pairs_by_set_from_response` emits all intra-group pairs, including that one.

Three entities with description-cosine edges (A,B)=0.70 and (B,C)=0.71 but (A,C)=0 — 2 of 3 pairs, above `MIN_DENSITY` — form one set. Both passes agree all three are one thing:

```
STATS {'entities': 3, 'candidates': 2, 'sets': 1, 'agreed_pairs': 3, 'merged': 2, ...}
MERGED GROUPS [['e1', 'e2', 'e3']]

last_pair_decisions -> 2 entries: (e2,e3) and (e1,e2)
e1/e3 reported? False
```

A and C were judged, twice, unanimously, and merged — and a caller enumerating `judge_pair_decisions` to review proposed merges will not see that merge at all.

This also means the docstring's justification doesn't hold for these pairs: "A pair no gate brought forward was never a question" is true of pairs outside every set, but (A,C) *was* put to the model and *was* answered. Iterating the union of `edges` and the judged pairs would cover it.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -893,5 +904,72 @@ def protected_idx(i: int, j: int) -> bool:
...
+                verdict = "split"
+            else:
+                verdict = "different"
+            decisions.append({
+                "a": ents[a]["id"],
+                "b": ents[b]["id"],
+                "a_name": ents[a].get("name") or "",
+                "b_name": ents[b].get("name") or "",
+                # What the gates measured, before the model was asked anything.
+                "similarity": round(float(similarity), 4),
```
galshubeli: **`similarity` is a constant for name-in-description pairs, not a measurement.**

`edges` is a mix of two different things. `knn_pairs` contributes measured cosines; `name_in_desc_pairs` contributes `NAME_IN_DESC_WEIGHT` — a flat 0.65 assigned to every pair it nominates, measuring nothing:

```python
out[(min(i, j), max(i, j))] = NAME_IN_DESC_WEIGHT
```

So a pair whose name and description cosines are both ~0.1, nominated only because A's name appears in B's description, publishes as `similarity: 0.65` — the same number as a pair with a genuine 0.65 cosine, under a field documented as "what the gates measured, before the model was asked anything."

This is the number the PR invites callers to form their bands from, and for these pairs it puts a measurement on the record that was never taken — the exact thing the "no probability, no score" section sets out to avoid. It also drives the sort, and therefore the cap.

Carrying the nomination source alongside (or leaving `similarity` null for name-in-desc pairs) would keep the field honest.

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
```diff
@@ -1358,7 +1363,9 @@ async def merge_group(members: list[dict[str, Any]]) -> tuple[str, set[str]]:
             logger.warning("LLM judge failed over the graph: %s", exc)
             # Merges committed before the failure are real; report them.
             self.last_judge_stats = dict(judge.last_stats)
+            self.last_judge_pair_decisions = list(judge.last_pair_decisions)
```
galshubeli, line 1366: **This line can only ever copy an empty list.**

`self.last_pair_decisions` is assigned at the very end of `LLMJudgeDeduplicator.deduplicate`, after the merge loop and the link loop. On the path this `except` handles, `_pair_decisions` never ran, so `judge.last_pair_decisions` is still the `[]` set at the top of the run.

Concretely: the judge merges three groups, then `self._merge` raises on the fourth. This handler correctly reports `merged: 3` from `judge.last_stats` — per the comment right above, "Merges committed before the failure are real; report them" — and then hands the caller `judge_pair_decisions: []` beside it. The stats claim merges the decision record has nothing to say about, and the line added here is dead.

Building the decisions incrementally, or computing them before the merge loop, would make the two agree.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -893,5 +904,72 @@ def protected_idx(i: int, j: int) -> bool:
...
+                "votes": votes,
+                "passes": 1 if unjudged else passes,
+                "verdict": verdict,
+                # Agreement is not the same as merger: a pair can be agreed and
+                # still kept apart by the grouping rules.
+                "merged": root_id.get(a) is not None and root_id.get(a) == root_id.get(b),
+            })
+            if len(decisions) >= _MAX_PAIR_DECISIONS:
```
galshubeli: **The cap keeps the most similar pairs, not the most consequential ones.**

Minor next to the others, but worth a thought. "the cut falls on the least similar" is true, and yet for a field whose stated purpose is putting a proposed merge in front of a reviewer, a merged pair is the last thing that should fall off the end.

It bites hardest because of the constant weight noted above: with more than 200 candidates and many name-in-description nominations all tied at exactly 0.65, `sorted` is stable, so which of those tied pairs survive the cut is decided by `edges` insertion order. A pair the judge agreed on and merged can be dropped while pairs the model rejected at the same 0.65 are kept.

Sorting by `(merged, verdict == "same", similarity)` before the cap would keep every decision that changed the graph.

galshubeli (commented):
Reviewed the per-pair decision record. The design reasoning is sound — reporting the parts rather than inventing a probability is the right call, and keeping `votes` separate from `merged` is correct.

The problem is that `_pair_decisions` reconstructs its facts from the judge's **answers** rather than from what the judge was **asked**. Two of the guarantees the PR description makes do not hold because of it, and both are reproducible end-to-end against the existing fixtures in `test_judge_dedup.py`. The 65 tests in that file pass; these paths simply aren't covered.

Five comments inline — the first two I'd treat as blocking, since they misreport merges and rejections in the one field built for a human to act on.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
Naseem77: Fixed in eb8fb0e.

`set_of_pair` is now built from `sets` before judging, so set membership is known from the sets themselves rather than inferred from what the model answered. The two passes' failed sets are tracked separately (`failed1` / `failed2`) instead of being unioned, and `passes` reports how many passes actually answered for that set.

So both passes failing now reads `verdict: unjudged`, `votes: 0`, `passes: 0`. Pass one answering and pass two failing reads `passes: 1`.

Three regression tests, all of which fail on the previous behaviour:

- `test_a_pair_the_judge_never_saw_is_not_reported_as_rejected` — `ScriptedLLM([None, None])`
- `test_a_pair_only_one_pass_saw_does_not_claim_two_passes` — `ScriptedLLM([[], None])`
- `test_an_unanswered_set_is_unjudged_rather_than_different` — a well-formed reply that partitions nothing, which is the `unanswered_sets` path rather than the failed-call one

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
Naseem77: Fixed in eb8fb0e. The loop is now over `set(edges) | set(set_of_pair)`, so every pair the model was asked about is recorded, not only the gate-nominated ones.

A pair reached only through its set carries `similarity: None` and `nominated_by: "set_expansion"` — it was a question, but nothing measured it, so no number is published for it.

The docstring was wrong for exactly the reason you gave, and now says so: every pair the model was asked about appears, which is not the same as every pair a gate nominated.

`test_a_pair_judged_only_through_its_set_is_still_recorded` builds your case — description cosines 0.70 and 0.71 with the third pair at 0, two of three pairs above the density floor — and asserts `candidates == 2`, one merged group of three, and that the unnominated pair is recorded as `same` / `merged`.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
Naseem77: Fixed in eb8fb0e, by leaving it null rather than carrying the source only.

`measured` is captured after the two `knn_pairs` gates and before `name_in_desc_pairs`, so `similarity` is published only where a cosine was actually taken. A pair nominated on `NAME_IN_DESC_WEIGHT` gets `similarity: None`, and a `nominated_by` field says which gate brought it forward: `embedding`, `name_in_desc` or `set_expansion`.

You were right that it also drove the sort and therefore the cap — that's addressed alongside your last comment.

`test_a_flat_nomination_weight_is_not_published_as_a_similarity` pins it with a pair linked only by name-in-description.

graphrag_sdk/src/graphrag_sdk/storage/deduplicator.py
Naseem77: Fixed in eb8fb0e. The merge and link loops are now wrapped in `try` / `finally`, with `_pair_decisions` called from the `finally` against `root_id` as it stands.

`root_id` was already built incrementally inside the merge loop, so a merge that raises part-way leaves it holding exactly the merges that were committed — the same ones `last_stats` reports. The two now agree on the failure path, and the line in `deduplicator.py` is no longer dead.

`test_merges_committed_before_a_failure_are_still_described` uses a merge hook that raises on its first call and asserts the judgement is still recorded (`same`, 2 of 2 passes) with `merged: False`, since nothing was committed for that pair.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
Naseem77: Fixed in eb8fb0e, with the key you suggested.

The list is now sorted before the cap on `(merged, verdict == "same", similarity, a, b)`, so a pair that changed the graph is never what falls off the end. An unmeasured `similarity` sorts last rather than ahead of a measured 0.99 on a missing value, and `a`/`b` break ties so the order no longer depends on `edges` insertion order — which was the other half of the problem you identified.

`test_decisions_are_ordered_by_similarity` asserted the old contract and would have been wrong to keep, so it's replaced by `test_the_cap_keeps_the_pairs_that_changed_the_graph`, which asserts merged pairs come first and similarity orders only within the remainder.

graphrag_sdk/src/graphrag_sdk/api/main.py
```diff
@@ -4913,6 +4913,9 @@ async def finalize(
             entities_linked=judge_stats.get("linked", 0),
             judge_llm_calls=judge_stats.get("llm_calls", 0),
             judge_stats=judge_stats,
+            judge_pair_decisions=list(
+                getattr(self._deduplicator, "last_judge_pair_decisions", []) or []
+            ),
```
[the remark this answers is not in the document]
Naseem77: Fixed in eb8fb0e. Added `test_the_per_pair_record_reaches_the_facade`, parametrized on `judge`, which reads `EntityDeduplicator.last_judge_pair_decisions` rather than the judge's own attribute — so it covers the copy across that boundary.

It asserts a populated decision comes back through the facade with the expected verdict, `merged`, and field set, and that the list stays empty when no judge LLM is passed.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -893,5 +904,72 @@ def protected_idx(i: int, j: int) -> bool:
...
+        forward was never a question.
+        """
+        passes = 2 if self._vote else 1
+        decisions: list[dict[str, Any]] = []
+        for (a, b), similarity in sorted(edges.items(), key=lambda kv: -kv[1]):
+            set_id = agreed.get((a, b), disagreed.get((a, b), same1.get((a, b), same2.get((a, b)))))
+            # A pair in a set whose prompt failed was judged fewer times than
+            # the run was configured for; saying "1 of 2 passes" about it would
+            # report a disagreement that never happened.
+            unjudged = set_id is not None and set_id in failed_sets
```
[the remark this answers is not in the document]
Naseem77: Fixed in eb8fb0e. `set_of_pair` — built from `sets` before judging — now resolves the set id independently of what the model answered, and each pass's failed sets are kept separate so `passes` counts the passes that actually answered.

A pair in a failed or unanswered set is therefore `unjudged` with the real pass count, not `different` with a full one. Covered by `test_a_pair_the_judge_never_saw_is_not_reported_as_rejected`, `test_a_pair_only_one_pass_saw_does_not_claim_two_passes` and `test_an_unanswered_set_is_unjudged_rather_than_different`.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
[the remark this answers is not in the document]
Naseem77: Fixed in eb8fb0e, as suggested. A candidate-pair-to-set-id map (`set_of_pair`) is built from `sets` before judging and used to resolve the set id, so the originating id survives for pairs that received no `SAME` vote.

Each pass's failed sets are also tracked separately rather than unioned, so `passes` reports how many passes actually answered rather than the configured count.

Regression test added for an unanswered set whose pairs receive zero `SAME` votes: `test_an_unanswered_set_is_unjudged_rather_than_different`, alongside two for the failed-call path.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -727,6 +734,10 @@ def protected(a: str, b: str) -> bool:
         def protected_idx(i: int, j: int) -> bool:
             return protected(ents[i]["id"], ents[j]["id"])
 
+        # Cleared here, not at the end: every early return below leaves a run
+        # with no decisions, and keeping the previous run's would attribute
+        # them to this one.
+        self.last_pair_decisions = []
```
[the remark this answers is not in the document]
Naseem77: Fixed in 2fb6efe. Both `last_pair_decisions` and `last_stats` are now cleared before `_fetch_entities`, so a fetch that raises leaves an empty record rather than the previous run's.

`last_stats` had the same flaw and is cleared alongside it — `EntityDeduplicator` reads both together on the failure path, and clearing only one would have made them disagree.

`test_a_reused_judge_does_not_carry_decisions_into_a_failed_run` covers it: one successful run, then the same instance against a graph whose entity fetch raises. It fails on the previous ordering.

In practice `EntityDeduplicator` builds a fresh judge per call, so the stale read was only reachable by using `LLMJudgeDeduplicator` directly — which is exported, so still worth closing.

On line 1044: that's the sort key in `_pair_decisions`, which has no reset in it. If a second site was meant, I couldn't find one — the only two assignments are the one moved here and the one in the `finally`, which is the run's real result.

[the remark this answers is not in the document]
Naseem77: Fixed in 2fb6efe. Both `last_pair_decisions` and `last_stats` are now cleared before `_fetch_entities`, so a fetch that raises leaves an empty record rather than the previous run's.

`last_stats` had the same flaw and is cleared alongside it — `EntityDeduplicator` reads both together on the failure path, and clearing only one would have made them disagree.

`test_a_reused_judge_does_not_carry_decisions_into_a_failed_run` covers it: one successful run, then the same instance against a graph whose entity fetch raises. It fails on the previous ordering.

In practice `EntityDeduplicator` builds a fresh judge per call, so the stale read was only reachable by using `LLMJudgeDeduplicator` directly — which is exported, so still worth closing.

On line 1044: that's the sort key in `_pair_decisions`, which has no reset in it. If a second site was meant, I couldn't find one — the only two assignments are the one moved here and the one in the `finally`, which is the run's real result.

(Apologies for the duplicate of this reply on an older thread — posted against a stale thread id.)

graphrag_sdk/src/graphrag_sdk/core/models.py
```diff
@@ -1010,6 +1010,38 @@ class FinalizeResult(DataModel):
     entities_linked: int = 0
     judge_llm_calls: int = 0
     judge_stats: dict[str, int | str] = Field(default_factory=dict)
+    judge_pair_decisions: list[dict[str, Any]] = Field(default_factory=list)
```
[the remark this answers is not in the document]
Naseem77: Added in 2fb6efe. `docs/api-reference.mdx:263` now lists `judge_pair_decisions` in the `FinalizeResult` return-field list, with the per-entry schema (`a`/`b`, `a_name`/`b_name`, `votes` out of `passes`, `verdict`, `merged`, `similarity`, `nominated_by`), the note that a pair reached only through its dense set is included though no gate nominated it, and both boundary behaviours: capped at 200 with merged pairs kept first, and empty when the judge did not run.

graphrag_sdk/tests/test_judge_dedup.py
```diff
@@ -1322,3 +1343,250 @@ def test_a_label_holding_the_separator_is_refused_and_a_recorded_one_reads_back_
...
+
+def test_the_cap_keeps_the_pairs_that_changed_the_graph():
+    graph = ScriptedGraph(ROWS)
+    _, dd, _ = _run_with_judge(graph, ScriptedLLM([[IBM], [IBM]]))
+
+    # The cap drops the tail, so the tail must be the pairs a reviewer least
+    # needs: a merge changed the graph and is never what falls off the end,
+    # and only then does similarity decide. Sorting on similarity alone would
+    # let a merged pair be cut while a rejected one at the same score stays.
+    keys = [(not d["merged"], d["verdict"] != "same") for d in dd.last_pair_decisions]
```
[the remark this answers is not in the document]
Naseem77: You're right — `ROWS` gives three pairs against a limit of 200, so that test asserted the ordering and nothing about the cap. Renamed to `test_decisions_are_ordered_by_consequence_not_similarity`, which is what it actually checks.

Added `test_the_cap_drops_the_tail_and_keeps_the_merge` in 2fb6efe, on a fixture built to make the drop consequential rather than just numerous:

- Sixty entities a degree apart on a circle. The top-k band nominates **316** candidate pairs, every one scoring above 0.99. Name vectors are zero so the name gate can't contribute.
- Two more sit off that plane at 0.56 to each other — over `DESC_GATE`, under everything else — and the judge calls those two the same thing.

So the run's single merge is the *least* similar pair in it. The test asserts `candidates == 316`, `len(decisions) == _MAX_PAIR_DECISIONS == 200`, that the merged pair is `decisions[0]`, and that its similarity is below every other kept pair's.

Verified against three counterfactuals, each failing the test: dropping the `[:_MAX_PAIR_DECISIONS]` slice, and removing either leading sort key — a cap taking the 200 most similar would discard the only merge and hand a reviewer 200 rejections with nothing to review.

graphrag_sdk/src/graphrag_sdk/storage/judge_dedup.py
```diff
@@ -840,58 +869,185 @@ def protected_idx(i: int, j: int) -> bool:
...
+        # ``last_stats`` reports them. A decision record assembled only on the
+        # way out would be empty beside a count of real merges, so the caller
+        # would be handed stats claiming merges the record says nothing about.
+        try:
+            for r, idxs in groups.items():
+                ids = [ents[i]["id"] for i in idxs]
+                if len(ids) < 2:
+                    root_id[idxs[0]] = ids[0]
+                    continue
+                surv, absorbed = await self._merge([byid[i] for i in ids])
```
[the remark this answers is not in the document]
Naseem77: Not applying this one — the suggested change would break the production path.

`self._merge` is the `merge_group` constructor argument. Its only real caller is `EntityDeduplicator._run_judge_phase`, which passes a plain `async def merge_group(members)` closure. A function has no `.merge` attribute, so `await self._merge.merge(...)` would raise `AttributeError` on every judged run.

The class the analyser points at, `FailingMerge` in the tests, defines `async def __call__`, which is callable — as does `RecordingMerge`, the hook every existing judge test uses. That one predates this PR (it's on `main` at `test_judge_dedup.py:358`), so the pattern isn't new here; the analyser picked up the new fixture as its example.

Covered by `test_agreed_pair_is_handed_to_the_merge_hook_once` and the rest of the merge-hook tests, plus the full suite at 2085 passed / 227 skipped.

